PR #7, opened 2014-07-14
Add base files for the branch (.travis.yml, .coveragerc, .gitignore, README.md, LICENSE)

guewen:

PR #8, opened 2014-07-14
Fix flake8 warnings and errors

guewen:
- [x] merge https://github.com/OCA/connector-magento/pull/7 before


guewen:
Depends of https://github.com/OCA/connector-magento/pull/7

guewen:
Waiting for https://github.com/OCA/connector-magento/pull/7 to be working

magentoerpconnect/unit/import_synchronizer.py
```diff
@@ -102,10 +104,11 @@ def _import_dependency(self, magento_id, binding_model,
                              class or parent class to use for the export.
                              By default: MagentoImportSynchronizer
         :type importer_cls: :class:`openerp.addons.connector.connector.MetaConnectorUnit`
```
guewen: Does someone know how I can reduce length of lines such as this one?
Is it possible to split sphinx references on 2 lines?

guewen:
I'm 2 lines close to be green, 2 lines that are > 80 and that I don't know how to split (sphinx cross-references)

guewen:
Green! :smile:

gurneyalex:
:+1:

pedrobaeza:
:+1:

bwrsandman:
:-1: 
Really not a fan of the liberal use of `#noqa` here. Most of these lines can be fixed.
The rest should be discussed as exceptions in travis' flake8 tests.

guewen:
I removed the # noqa at end of lines.
I won't remove the global # flake8: noqa because they are appropriate:
- conf.py is a sphinx generated configuration file
- test_data\* are test data automatically recorded from RPC requests and are thousands lines long, it would be ridiculous to correct pep8 in them

@bwrsandman  I don't get what you mean by 

> The rest should be discussed as exceptions in travis' flake8 tests.

bwrsandman:
@guewen for example, if  magentoerpconnect/doc/conf.py has to ignore a certain rule, we can add the filename pattern to the exceptions list in the quality tools the same way we ignore unused imports in **init**.py.

For the remaining noqa, can you add a comment explaining why they're there. Such as: 

> E00 needs to be ignored.

guewen:
Again, conf.py is a configuration file, which lies to be also a Python file. It is generated by Sphinx but is not generated being pep8 valid (for instance, all the commented options miss a space after #, eg: `#needs_sphinx = '1.0'`). If a convention exists that says that the doc is always in module/doc, we could add a rule in the quality tools, but I don't think such a convention actually exists. Meantime, I am not sure to get exactly where the evil is to use `# flake8: noqa` on a file such as this one.

For the other files where there is a `# flake8: noqa`, **all** the rules must be ignored, because they are not regular code, only dicts of values of thousands of lines recorded for the tests (they could have been text files as well but would have required a useless parsing then). I can add a comment to say that.

bwrsandman:
Forgive my ignorance of sphinx generated python, but is there a reason to commit it if it's automatically generated?

If the file is always called `conf.py`, we can tell test_flake8 to ignored the file pattern `conf.py`

As I explained elsewhere, we should be avoiding `#noqa` for the following reasons:
- You're skipping all tests while you may just want to ignore a certain rule or filename
- The style will suffer
- It gives an example to other devs that they don't really need to follow PEP8
- As a reviewer you have no idea why `#noqa` is used unless a comment is added
- PEP8 does other checks and raises syntax errors and the like

guewen:
The `conf.py` file is generated at first, then needs to be customized with the options of the project, so it should be commited. Sphinx scaffolds the doc.

I totally understand the reasons to avoid `# noqa` but we are speaking of files that are not strictly part of the code, these files could as well have been `cfg` files for `conf.py` or `json` or `txt` files for the recorded data.

guewen:
Maybe I could do a test with autopep8 on the files with the recorded data, but noone can't do that manually anyway.

guewen:
> If the file is always called conf.py, we can tell test_flake8 to ignored the file pattern conf.py

I think that worse because it could ignore a file that could be part ouf a module with real code in it.
But perhaps could we generalize the `module/doc/conf.py` pattern.

guewen:
autopep8 won't improve the long lines in the data files

guewen:
I don't think I can do more, I ran `autopep8` on the data files but it doesn't cut the long lines.
@bwrsandman can you check again please? I would like this PR to be merged since it is a prerequisite for all the other PR of the branch.

magentoerpconnect/doc/conf.py
```diff
@@ -11,6 +11,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+# flake8: noqa
```
bwrsandman: Please add an explanation.

guewen: @bwrsandman fixed pep8 in the sphinx config file

bwrsandman:
:+1: Thank you @guewen

PR #11, opened 2014-07-14
Update of Magento stock quantity: reduced the memory footprint and improved...

guewen:
performance by using read() instead of browse() and proceeding by steps of 1000
records at a time.

Bug report: https://code.launchpad.net/bugs/1330450
Original MP: https://code.launchpad.net/~camptocamp/openerp-connector-magento/7.0-update-stock-1330450/+merge/223393

### DIFF
--- a/magentoerpconnect/product.py
+++ b/magentoerpconnect/product.py
@@ -25,6 +25,7 @@ import urllib2
 import base64
 import xmlrpclib
 import sys
+from collections import defaultdict
 from openerp.osv import orm, fields
 from openerp.tools.translate import _
 from openerp.addons.connector.queue.job import job, related_action
@@ -51,6 +52,11 @@ from .backend import magento
 from .related_action import unwrap_binding
 
 _logger = logging.getLogger(__name__)
+
+
+def chunks(items, length):
+    for index in xrange(0, len(items), length):
+        yield items[index:index + length]
 
 
 class magento_product_product(orm.Model):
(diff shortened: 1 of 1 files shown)


guewen:
Rebased

magentoerpconnect/product.py
```diff
@@ -123,35 +129,86 @@ def _product_type_get(self, cr, uid, context=None):
...
-        return product_stk[stock_field]
+        location_ctx['location'] = location.id
+
+        product_fields = ['magento_qty', stock_field]
+        if read_fields:
+            product_fields += read_fields
+
+        product_ids = [product['id'] for product in products]
+        for chunk_ids in chunks(product_ids, self.RECOMPUTE_QTY_STEP):
+            for product in self.read(cr, uid, chunk_ids, product_fields,
```
gurneyalex, line 188: I may be missing something obvious here, but would there not be a way of avoiding the 2 calls to read() by passing a slightly longer list of fields in the call on line 150?

magentoerpconnect/product.py
```diff
@@ -123,35 +129,86 @@ def _product_type_get(self, cr, uid, context=None):
...
         return True
 
-    def _magento_qty(self, cr, uid, product, context=None):
+    def _recompute_magento_qty_backend(self, cr, uid, backend, products,
+                                       read_fields=None, context=None):
+        """ Recompute the products quantity for one backend.
+
+        If field names are passed in ``read_fields`` (as a list), they
+        will be read in the product that is used in
+        :meth:`~._magento_qty`.
```
gurneyalex, line 167: this value is not passed by the call to the method on line 157, so I don't understand how an extended  `_magento_qty` method can use additional fields.

guewen, line 167: The call on 157 does not need additional fields so it does not use this attribute.
If one want to pass additional fields, it can inherit the model and something along these lines:

```
def _recompute_magento_qty_backend(self, cr, uid, backend, products, 
                                                                 read_fields=None, context=None):
    return super(magento_product_product, self)._recompute_magento_qty_backend(
        self, cr, uid, backend, products, read_fields=['field1', 'field2'], context=context)
```

magentoerpconnect/product.py
guewen, line 188: The second call is made in the context of a location (with a location in the context) in order to get their stock quantities. We need to read the products once to get their backend from which we find their location, then read them again with the location in the context.

nbessi:
:+1:

OSguard:
:+1:

sebastienbeau:
Good Work Guewen :+1:

PR #17, opened 2014-07-17
Add a 'to date' boundary to batch jobs

guewen:
Currently, the batch jobs have a 'from date' boundary only, which is
incremented each time a new batch job is created.

This may cause an issue when many batch jobs are created and don't have
the time to be processed: the same records will be imported several
times.

Example with 4 batch jobs with an interval of 10 minutes (the 30 seconds
difference is a buffer because Magento may commit a new record 'created_at' with a
delay) :

  Batch job 1 from 2014-07-17 10:00:00
  Batch job 2 from 2014-07-17 10:09:30
  Batch job 3 from 2014-07-17 10:19:00
  Batch job 4 from 2014-07-17 10:28:30

A product A is modified on Magento at 2014-07-17 10:25:00
None of the batch job could been executed so far (maybe an interruption
or a high load).  When the batch job 1 will be processed, it will see
that product A has been modified later than 10:00 so it creates an
import job. Turn of batch job 2, it sees the product modified after
10:10, creates an import job... and so on.

This patch adds a 'to' boundary so the batch jobs become:

  Batch job 1 from 2014-07-17 10:00:00 to 2014-07-17 10:10:00
  Batch job 2 from 2014-07-17 10:09:30 to 2014-07-17 10:19:30
  Batch job 3 from 2014-07-17 10:19:00 to 2014-07-17 10:19:00
  Batch job 4 from 2014-07-17 10:28:30 to 2014-07-17 10:28:30

Thus, in the same situation than before, only the batch job 3 would see
the record.

### DIFF
--- a/magentoerpconnect/product.py
+++ b/magentoerpconnect/product.py
@@ -190,17 +190,21 @@ class ProductProductAdapter(GenericAdapter):
             else:
                 raise
 
-    def search(self, filters=None, from_date=None):
-        """ Search records according to some criterias
+    def search(self, filters=None, from_date=None, to_date=None):
+        """ Search records according to some criteria
         and returns a list of ids
 
         :rtype: list
         """
         if filters is None:
             filters = {}
+        dt_fmt = '%Y/%m/%d %H:%M:%S'
         if from_date is not None:
-            str_from_date = from_date.strftime('%Y/%m/%d %H:%M:%S')
-            filters['updated_at'] = {'from': str_from_date}
+            filters.setdefault('updated_at', {})
+            filters['updated_at']['from'] = from_date.strftime(dt_fmt)
+        if to_date is not None:
+            filters.setdefault('updated_at', {})
+            filters['updated_at']['to'] = to_date.strftime(dt_fmt)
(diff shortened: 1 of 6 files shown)


pedrobaeza:
:+1: (code review)

guewen:
Rebased

gurneyalex:
:+1:

magentoerpconnect/product.py
```diff
@@ -190,17 +190,21 @@ def _call(self, method, arguments):
...
-        """ Search records according to some criterias
+    def search(self, filters=None, from_date=None, to_date=None):
+        """ Search records according to some criteria
         and returns a list of ids
 
         :rtype: list
         """
         if filters is None:
             filters = {}
+        dt_fmt = '%Y/%m/%d %H:%M:%S'
```
bwrsandman: Can this be factored out as a global variable in `__init__.py` , maybe?

guewen: You are right, done (put in backend_adapter.py which is where the things related to the Magento API lie).

bwrsandman:
:+1:

PR #19, opened 2014-07-23
[IMP] Make compatible with php 5.4

albertofl:


magentoerpconnect/unit/backend_adapter.py
```diff
@@ -152,7 +152,12 @@ def _call(self, method, arguments):
             with magentolib.API(self.magento.location,
                                 self.magento.username,
                                 self.magento.password,
-                                full_url=custom_url) as api:
+                                full_url=custom_url) as api:                
+                # If magento is used with PHP version >= 5.4, then we have to remove all
+                # trailing None values
+                if isinstance(arguments, list):
+                    while arguments and arguments[-1] is None:
+                        arguments.pop()
```
bwrsandman, line 160: Loop popping a list is very inefficient.
If you're expecting a long list try this which is O(N):

``` python
i = len(arguments)
while i > 0 arguments[i-1] is None:
    i -= 1
arguments = arguments[:i]
```

The thing is, when you pop or append from a list, internally the list will eventually resize, which means copying the contents to another place in memory. Because of this, doing it in a loop is a problem because it could resize N times giving it a performance of O(N2) or O(NlogN)

guewen, line 160: In that case, the list will not be very long because that's the list of arguments for the Magento's API methods. Most of the time, there is 1 to 3 arguments, sometimes a few more. When popping, the list is shrunk when the new size is less than the half of the allocated size, which will never or hardly never happen here since usually there is only 1 or 2 trailing None, if any. So I prefer the readability of the popping in this specific case.

bwrsandman, line 160: OK

bwrsandman:
:+1:

magentoerpconnect/unit/backend_adapter.py
gurneyalex, line 160: @bwrsandman: actually, append / pop (without an index, iow popping from the end of a Python list) is very efficient in Python with a complexity ~O(1) (amortized for append()). insert() and pop from anywhere else than the end is O(n) and should be avoided.

gurneyalex:
:+1:

guewen:
:+1:

guewen:
Mmh no, tests no longer pass :-1:

guewen:
I opened https://github.com/OCA/connector-magento/pull/28 where I fixed the tests.
So I close this PR. Thanks for the fix.

PR #21, opened 2014-07-23
[IMP] Changes in total_amount and create payment when the store currecy ...

albertofl:
[IMP] Changes in total_amount and create payment when the store currecy is different to sale order currency

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -465,7 +465,12 @@ class SaleOrderImport(MagentoImportSynchronizer):
         if not mag_sale.payment_method_id.journal_id:
             return
         sale_obj = sess.pool['sale.order']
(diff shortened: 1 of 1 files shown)


guewen:
Can you rebase please?

OSguard:
i think this change is incomplete and needs test.

I miss the handling of SaleOrderLineImportMapper including discount. I think only handling total will not cover all case and there will be need for some addational work.
As well it only will work if the base currency in magento is the same as in OpenERP. So we need a additional description.

But i m not a export on multi currency setup.

pedrobaeza:
@liebana, are you going to work on this?

pedrobaeza:
As there's no answer, I close. Feel free to reopen in case there's interest.

PR #24, opened 2014-08-05
[IMP] add the posibility to merge the export job, this will avoid useles...

sebastienbeau:
...s call to Magento

Please test it with this branch of connector : https://github.com/akretion/connector/compare/OCA:7.0...7.0_merge_job?expand=1

Test are OK

### DIFF
--- a/magentoerpconnect/consumer.py
+++ b/magentoerpconnect/consumer.py
@@ -33,6 +33,19 @@ from .connector import get_environment
 _MODEL_NAMES = ()
 _BIND_MODEL_NAMES = ()
 
+def _job_to_key(job):
+    return '%s-%s-%s-%s' % (
+        job.func_name,
+        ', '.join([repr(arg) for arg in job.args]),
+        job.company_id,
+        job.user_id,
+        )
+
+def _merge_job(existing_job, job):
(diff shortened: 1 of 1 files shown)


guewen:
Can you give a use case where the same job is generated twice in the same transaction?

If this is really needed, it should be done automatically without needing to pass the 2 functions (that seems generic).

sebastienbeau:
Hi a real and simple case is for exemple if you export the product. You will
- add a consummer on_write on the model product_product
- add a consummer on_write on the model product_template

If you write on the product template directly one job will be created => OK
If you write on the product_product 2 job will be created as the write will be trigger on the product_product and on the product_template

An other solution will to trigger only the field of the current model. But It maybe not fit with all of the case (for exemple if an extra module call a new write after the first one, bad practice but sometime we do not have the choice).
Also I am loooking to support store function field. And in this case I think we will need the merge feature because if I have 2 function field that depend on a category and save my category I do not want to create one job for each field.

I think also that the code is really generic and I use for all magento connector export without trouble. So I can directly implement it in the connector.

guewen:
Yes it should be generic and it should not need code in the implementations of the connector.

PR #9, opened 2014-07-14
When a sales order is canceled, when one uses the 'New Copy from Quotati...

guewen:
...on' button,

the binding are moved from the old (canceled) order to the new one, so the
synchronizations now happen on the new order.
- [x] **Warning**: needs this PR to be merged in _addons_ https://github.com/odoo/odoo/pull/1025
- [x]  **Warning**:  needs the same fix in OCB

Original MP on Launchpad: https://code.launchpad.net/~camptocamp/openerp-connector-magento/7.0-copy-quotation-move-binding/+merge/226137

PR #25, opened 2014-08-13
Export cancellation of sales orders to Magento

guewen:
When a sales order is cancelled on Odoo, a job to export the sales
order on Magento is generated.

When the button to create a new sales order from the canceled one
is used on Odoo, the sales order on Magento is reopened.

The job does not use the 'sales_order.cancel' Magento API method
because this method prevents to reopen the order later (it changes
the 'state' field). Instead, it uses the 'sales_order.addComment' that
change the 'status' of the order and allow to reopen the status later.
- [x] Needs #9 to be merged before


guewen:
The diff includes #9 until it is merged.

guewen:
Travis is green on odoo and red on OCB because OCB lacks this fix: odoo/odoo#1025

guewen:
Rebased

OSguard:
We add new Button "cancel sale.order" which opens a popup to choose "cancel as well in Magento" or only in "OpenERP" so you have more control. So then we can decide if we use the cancel api call and not only setting a status and the Magento state is wrong.
May we should make this module work with this Pull-Request.

guewen:
The owners of the Magento shop who asked us this feature often needs to reopen canceled orders on the request of their customers, that's why we only put the status to canceled and not the state. In your case, when they would have to click on the button, they wouldn't know if they would have to reopen it later or not. Users of the shop still seee the orders as canceled even if the state is not canceled so that's seems to be alright, besides, it was already implemented with the 'status' in the previous implementations of the connector (5 to 6.1) (I don't know if leaving the orders' status as canceled with a different state as side effects though).

OSguard:
Resurrection (cancel => alive) is always a mess up. The point is why you would trigger the cancel event is to re enable the coupon code use for this order. (you need a fix on magento side as well). But this makes revers cancel impossible.
So mixin both solution do not work. So maybe a option solution? We can add a option "use wizard".

guewen:
They also want to revive a sale order to keep the payment authorization if I remember well

sebastienbeau:
:+1:  I totaly approve the approche, It really cool to be able to re-open it from Odoo. Thanks

mdietrichc2c:
:+1: (Travis error is unrelated to the connector)

bwrsandman:
Restarting build. 7.0 branch is green, so this could be good.

guewen:
Rebased

guewen:
Fixed the tests, it _should_ be green

guewen:
Green!

yvaucher:
:+1:

PR #28, opened 2014-08-15
Remove trailing None on calls made to the Magento's API (PHP 5.4 compatibility)

guewen:
When Magento is used with PHP 5.4+, the API may return errors if we send
API calls with None in the last arguments. So, we remove all the trailing
None from the arguments before sending.


guewen:
Superseed https://github.com/OCA/connector-magento/pull/19

yvaucher:
@guewen To be rebased

guewen:
Rebased

gurneyalex:
:+1:

nbessi:
:+1:

yvaucher:
:+1:

PR #33, opened 2014-09-10
Add mechanisms for the export of records

guewen:
- Prevent concurrent exports to run on the same binding record by acquiring a
  lock on the binding record before exporting it
- Add a 'export_dependency' helper method in the Exporter to ease the export
  of dependencies

PR #36, opened 2014-09-11
7.0 export partner and address

florian-dacosta:
Completion of the magentoerpconnect_export_partner module to fix the export of partners and add the export of addresses.

This PR depend of this one and need to be merged after : 
- [x] https://github.com/OCA/connector-magento/pull/33

Initial proposal on lauchpad : 
https://code.launchpad.net/~akretion-team/openerp-connector-magento/7.0-export-partner-and-address/+merge/224180

### DIFF
--- a/magentoerpconnect/partner.py
+++ b/magentoerpconnect/partner.py
@@ -79,7 +79,7 @@ class magento_res_partner(orm.Model):
     _inherits = {'res.partner': 'openerp_id'}
     _description = 'Magento Partner'
 
-    _rec_name = 'website_id'
+    _rec_name = 'name'
 
     def _get_mag_partner_from_website(self, cr, uid, ids, context=None):
         mag_partner_obj = self.pool['magento.res.partner']
@@ -190,6 +190,8 @@ class magento_address(orm.Model):
     _sql_constraints = [
         ('magento_uniq', 'unique(backend_id, magento_id)',
          'A partner address with same ID on Magento already exists.'),
+        ('openerp_uniq', 'unique(backend_id, openerp_id)',
+         'A partner address can only have one binding by backend.'),
     ]
 
 
@@ -558,6 +560,12 @@ class AddressAdapter(GenericAdapter):
         return [int(row['customer_address_id']) for row
                 in self._call('%s.list' % self._magento_model,
                               [filters] if filters else [{}])]
+
+    def create(self, data):
+        """ Create a record on the external system """
+        partner_id = data.pop('partner_id')
+        return self._call('%s.create' % self._magento_model,
+                          [partner_id, data])
 
 
 @magento
--- a/magentoerpconnect/partner.py
+++ b/magentoerpconnect/partner.py
@@ -561,11 +561,10 @@ class AddressAdapter(GenericAdapter):
                 in self._call('%s.list' % self._magento_model,
                               [filters] if filters else [{}])]
 
-    def create(self, data):
+    def create(self, customer_id, data):
         """ Create a record on the external system """
-        partner_id = data.pop('partner_id')
         return self._call('%s.create' % self._magento_model,
-                          [partner_id, data])
+                          [customer_id, data])
 
 
 @magento
--- a/magentoerpconnect_export_partner/partner.py
+++ b/magentoerpconnect_export_partner/partner.py
@@ -48,11 +48,11 @@ class PartnerExport(MagentoExporter):
             'magento_partner_id': self.binding_id,
         }
         if not self.binding_record.magento_address_bind_ids \
-        and not self.binding_record.consider_as_company:
+        and not self.binding_record.consider_as_company \
+        and self.binding_record.street:
             data['openerp_id'] = self.binding_record.openerp_id.id
-
+            data['is_default_billing'] = True
             if not self.binding_record.child_ids:
-                data['is_default_billing'] = True
                 data['is_default_shipping'] = True
 
             with self._retry_unique_violation():
@@ -203,6 +203,7 @@ class PartnerAddressExportMapper(ExportMapper):
     @changed_by('street', 'street2')
     @mapping
     def street(self, record):
+        street = False
         if record.street:
             street = record.street
         if record.street2:
--- a/magentoerpconnect_export_partner/partner.py
+++ b/magentoerpconnect_export_partner/partner.py
@@ -47,30 +47,34 @@ class PartnerExport(MagentoExporter):
         data = {
             'magento_partner_id': self.binding_id,
         }
-        if not self.binding_record.magento_address_bind_ids \
-        and not self.binding_record.consider_as_company \
-        and self.binding_record.street:
-            data['openerp_id'] = self.binding_record.openerp_id.id
-            data['is_default_billing'] = True
+        # Condition on street field because we consider that if street
+        # is false, there is no address to export on this partner.
+        if (not self.binding_record.magento_address_bind_ids and
+                not self.binding_record.consider_as_company and
+                self.binding_record.street):
+            extra_vals = {
+                'is_default_billing': True,
+            }
             if not self.binding_record.child_ids:
-                data['is_default_shipping'] = True
-
-            with self._retry_unique_violation():
-                self.session.create('magento.address', data)
+                extra_vals['is_default_shipping'] = True
+            self._export_dependency(self.binding_record.openerp_id,
+                                    'magento.address',
+                                    exporter_class=AddressExport
+                                    binding_field='magento_address_bind_ids',
+                                    binding_extra_vals=extra_vals)
 
         for child in self.binding_record.child_ids:
+            child_extra_vals = {}
             if not child.magento_address_bind_ids:
                 if child.type == 'invoice':
-                    data['is_default_billing'] = True
-                else:
-                    data['is_default_billing'] = False
+                    child_extra_vals['is_default_billing'] = True
                 if child.type == 'delivery':
-                    data['is_default_shipping'] = True
-                else:
-                    data['is_default_shipping'] = False
+                    child_extra_vals['is_default_shipping'] = True
                 data['openerp_id'] = child.id
-                with self._retry_unique_violation():
-                    self.session.create('magento.address', data)
+                self._export_dependency(child, 'magento.address',
+                                        exporter_class=AddressExport
+                                        binding_field='magento_address_bind_ids',
+                                        binding_extra_vals=child_extra_vals)
 
     def _validate_create_data(self, data):
         """ Check if the values to import are correct
@@ -92,8 +96,8 @@ class AddressExport(MagentoExporter):
 
     def _export_dependencies(self):
         """ Export the dependencies for the record"""
-        relation = self.binding_record.parent_id or \
-            self.binding_record.openerp_id
+        relation = (self.binding_record.parent_id or
+                    self.binding_record.openerp_id)
         self._export_dependency(relation, 'magento.res.partner',
                                 exporter_class=PartnerExport)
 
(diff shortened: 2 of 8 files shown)


magentoerpconnect_export_partner/partner.py
```diff
@@ -34,27 +35,106 @@
...
 class PartnerExport(MagentoExporter):
     _model_name = ['magento.res.partner']
 
+    def _after_export(self):
+        data = {
+            'magento_partner_id': self.binding_id,
+        }
+        if not self.binding_record.magento_address_bind_ids \
+        and not self.binding_record.consider_as_company \
+        and self.binding_record.street:
```
guewen: This indentation is better (avoid `\`, indent below `if`):

```
if (not self.binding_record.magento_address_bind_ids and
        not self.binding_record.consider_as_company and
        self.binding_record.street):
```

BTW why is there a condition on `self.binding_record.street`?

magentoerpconnect_export_partner/partner.py
```diff
@@ -34,27 +35,106 @@
...
+            if not self.binding_record.child_ids:
+                data['is_default_shipping'] = True
+
+            with self._retry_unique_violation():
+                self.session.create('magento.address', data)
+
+        for child in self.binding_record.child_ids:
+            if not child.magento_address_bind_ids:
+                if child.type == 'invoice':
+                    data['is_default_billing'] = True
```
guewen: you are reusing the same `data` dict which seems a bit risky, I would prefer if you'd worked with copies.

magentoerpconnect_export_partner/partner.py
```diff
@@ -34,27 +35,106 @@
...
+        if not self.binding_record.magento_address_bind_ids \
+        and not self.binding_record.consider_as_company \
+        and self.binding_record.street:
+            data['openerp_id'] = self.binding_record.openerp_id.id
+            data['is_default_billing'] = True
+            if not self.binding_record.child_ids:
+                data['is_default_shipping'] = True
+
+            with self._retry_unique_violation():
+                self.session.create('magento.address', data)
```
guewen: Is there a reason why you do not use `self._export_dependency()`?

magentoerpconnect_export_partner/partner.py
```diff
@@ -34,27 +35,106 @@
...
+                    data['is_default_billing'] = True
+                else:
+                    data['is_default_billing'] = False
+                if child.type == 'delivery':
+                    data['is_default_shipping'] = True
+                else:
+                    data['is_default_shipping'] = False
+                data['openerp_id'] = child.id
+                with self._retry_unique_violation():
+                    self.session.create('magento.address', data)
```
guewen: Is there a reason why you do not use `self._export_dependency()`?

magentoerpconnect_export_partner/partner.py
```diff
@@ -34,27 +35,106 @@
...
+
+
+@magento
+class AddressExport(MagentoExporter):
+    _model_name = ['magento.address']
+
+    def _export_dependencies(self):
+        """ Export the dependencies for the record"""
+        relation = self.binding_record.parent_id or \
+            self.binding_record.openerp_id
```
guewen: Please avoid `\`, use parenthesis

magentoerpconnect_export_partner/partner.py
```diff
@@ -68,3 +148,65 @@ def names(self, record):
...
+              ('company', 'company'),
+              ]
+
+    @changed_by('parent_id', 'openerp_id')
+    @mapping
+    def partner(self, record):
+        binder = self.get_binder_for_model('magento.res.partner')
+        erp_partner_id = record.parent_id.id \
+            if record.parent_id \
+            else record.openerp_id.id
```
guewen: Use parenthesis instead of \ or a normal if: else:

magentoerpconnect_export_partner/tests/test_synchronization.py
```diff
@@ -0,0 +1,123 @@
...
+            self.assertEqual(len(calls_done), 1)
+
+            method, [partner, values] = calls_done[0]
+            self.assertEqual(method, 'customer_address.create')
+            self.assertEqual(partner, 1)
+            self.assertEqual(values['postcode'], '00000')
+            self.assertEqual(values['city'], 'City test')
+            self.assertEqual(values['street'], '15, test street')
+            self.assertEqual(values['country_id'], 'BE')
+            self.assertEqual(values['telephone'], '0000000000')
```
guewen: Great, thank you very much for the tests!

magentoerpconnect/partner.py
```diff
@@ -559,6 +561,12 @@ def search(self, filters=None):
                 in self._call('%s.list' % self._magento_model,
                               [filters] if filters else [{}])]
 
+    def create(self, data):
```
guewen: I  would prefer if you modify the arguments of `create()` for this adapter like this `def create(self, customer_id, data)`. I think we would gain in clarity.

magentoerpconnect_export_partner/partner.py
florian-dacosta: Ok, i'll change that.
I add the condition on the street to avoid to try exporting an adress for a partner if it does not have any address. (Would create a job in fail for nothing)
I think we can assume that if street is present, then the partner has an address (so we want to try creating one in magento), if it isn't present, it does not have any address and we do not even want to try creating an address on magento.
I put a condition on street, but it could have been on country_id or city...
What do you think?

guewen: Understood. Can you add a comment in the code to say that?

magentoerpconnect_export_partner/partner.py
florian-dacosta: I am not sure, I thought it was meant to do the other way : If I export an address, I first export the partner using `self._export_dependencies()`. 
But in my case, I export a partner, and then I want to export the linked address if any. It is the opposite of dependency no?

I took a look at `def _export_dependencies`, and even if I is right to use it in my case, after an export, I am not sure it is appropriate because the magento.address take different mandatory field than backend_id but it takes magento_partner_id.
So if I give my address record to self._export_dependency(), it won't create binding properly.

But I may miss an important point here!

magentoerpconnect_export_partner/partner.py
florian-dacosta: Of course!
I'll take into account all your comment once we have resolved the export_dependency point so I won't change everything twice!

magentoerpconnect_export_partner/partner.py
guewen: > But in my case, I export a partner, and then I want to export the linked address if any. It is the opposite of dependency no?

Yes but nothing prevents you to call _export_dependency() in _after_create().

> I took a look at def _export_dependencies, and even if I is right to use it in my case, after an export, I am not sure it is appropriate because the magento.address take different mandatory field than backend_id but it takes magento_partner_id.

So you can add a argument with values to set in the binding record.

florian-dacosta: Ok. The thing is, as my branch depend of the PR 33 where the export_dependency() is defined, I can't modify the method in this branch.
I'am gonna make a new one for this feature to merge in your branch 7.0-export-dependency+lock then.

guewen: You don't need to create a new one. Use `git rebase guewen/7.0-export-dependency+lock` (do not forget to add the git remote "guewen").

florian-dacosta: Well, there is one last point I don't get : in _export_dependency, in order to export there is a condition : 

```
rel_binder = self.get_binder_for_model(binding_model)
if rel_binder.to_backend(binding_id) is None:
```

In my tests, after creating the bindings for my addresses, it does not pass this condition, and so, it won't export my addresses.
I just don't get what this is... not what should I do in order to make it pass the condition

florian-dacosta: The problem is that rel_binder.to_backend(binding_id) return False, because it is a new binding.
But it does not return None. So I am wondering if we could change the condition by : 
if not rel_binder.to_backend(binding_id):

guewen: I think it can be replaced by this condition, however it should return None and not False: https://github.com/OCA/connector-magento/blob/7.0/magentoerpconnect/unit/binder.py#L104

florian-dacosta: Well, here, `if not rel_binder.to_backend(binding_id):`
We do not pass the wrap argument because we pass the binding id, so it won't return None. Even if we passed it, it wouldn't because
We just create the binding in the export_dependency method so it exists.

The to_backend method return  `magento_record['magento_id']` which is False because the address is not yet created in magento.

I am not sure to understand exactly the objective of the condition, but I guess the goal is to export only if the record is not on magento yet. In this case, we expect the  `rel_binder.to_backend(binding_id)` to return False.

magentoerpconnect/partner.py
florian-dacosta: For this part I think it is too complitcated because I would have to also modify `_create()` so it call `create()` with the right arguments, and I would have to rewrite entirely `_run()` so it call `_create()` with the right arguments. I am not sure it worth it.

guewen: But I think you could pop the 'partner_id' from the data right in exporter._create() instead of in the adapter so the public interface of the adapter is explicit. Only an idea... maybe not the best one. It seems to me that the sooner we can extract the partner id the better it would be because it means less steps to remember the "where does comes this partner_id and why is it in the data".

florian-dacosta: Ok, I get your point!

magentoerpconnect_export_partner/partner.py
```diff
@@ -68,3 +159,66 @@ def names(self, record):
...
+    @changed_by('parent_id', 'openerp_id')
+    @mapping
+    def partner(self, record):
+        binder = self.get_binder_for_model('magento.res.partner')
+        if record.parent_id:
+            erp_partner_id = record.parent_id.id
+        else:
+            erp_partner_id = record.openerp_id.id
+        mag_partner_id = binder.to_backend(erp_partner_id, wrap=True)
+        return {'partner_id': mag_partner_id}
```
guewen, line 184: `partner_id` lets think that it is the openerp partner id, `customer_id` would let us guess that we speak about the magento customer

guewen:
Could you rebase before the merge? (and ideally change the 'partner_id' to 'customer_id')

guewen:
@florian-dacosta I rebased locally in the idea to merge, but the tests give:

```
test_1_export_partner (openerp.addons.magentoerpconnect_export_partner.tests.test_synchronization.TestMagentoPartnerExport)
Test export of partner ... ERROR
test_2_export_partner_address (openerp.addons.magentoerpconnect_export_partner.tests.test_synchronization.TestMagentoPartnerExport)
Test export of address ... ERROR

======================================================================
ERROR: test_1_export_partner (openerp.addons.magentoerpconnect_export_partner.tests.test_synchronization.TestMagentoPartnerExport)
Test export of partner
----------------------------------------------------------------------
  File "/home/guewen/code/camptocamp/magento7/parts/connector-magento/magentoerpconnect_export_partner/tests/test_synchronization.py", line 84, in test_1_export_partner
    partner_helper = self.get_magento_helper('magento.res.partner')
AttributeError: 'TestMagentoPartnerExport' object has no attribute 'get_magento_helper'

======================================================================
ERROR: test_2_export_partner_address (openerp.addons.magentoerpconnect_export_partner.tests.test_synchronization.TestMagentoPartnerExport)
Test export of address
----------------------------------------------------------------------
  File "/home/guewen/code/camptocamp/magento7/parts/connector-magento/magentoerpconnect_export_partner/tests/test_synchronization.py", line 108, in test_2_export_partner_address
    partner_helper = self.get_magento_helper('magento.res.partner')
AttributeError: 'TestMagentoPartnerExport' object has no attribute 'get_magento_helper'

----------------------------------------------------------------------
```

guewen:
Cherry-picked https://github.com/akretion/connector-magento/commit/609e7b26136890cc9ece00bb81737812b6c367c1 and https://github.com/akretion/connector-magento/commit/609e7b26136890cc9ece00bb81737812b6c367c1

guewen:
Merged in 49dde19c7bfa9a49a74f27795f2218d0d96f2266

florian-dacosta:
Hello Guewen,

Sorry, I just saw your comments now.
I am a little confused,  it seems that you merged it, but the PR is still open and has conflicts.
Should I still do something?
For the tests, as you saw, it did depend of another PR.

guewen:
@florian-dacosta I merged the missing commits, resolved the conflicts, issues and the tests, I just miss a few things and that broke the build, fixed here: https://github.com/OCA/connector-magento/pull/83

GitHub did not show the PR as merged because I merged it with git directly.

PR #38, opened 2014-09-17
[FIX] products search before api call: manage case where from_date is in...

### LINKED ISSUE #37: wrong format date in products search just before calling catalog_product.list
in connector-magento / magentoerpconnect / product.py

<pre>
    def search(self, filters=None, from_date=None):
        """ Search records according to some criterias
        and returns a list of ids

        :rtype: list
        """
        if filters is None:
            filters = {}
        if from_date is not None:
            str_from_date = from_date.strftime('%Y/%m/%d %H:%M:%S')
            filters['updated_at'] = {'from': str_from_date}
</pre>

it seems here that from_date information is in filters, and not in from_date arg.
Then the date is not well formatted according Magento needs and Magento returns an error.

rdeheele:
... filters
fixes #37 
it seems here that from_date information can be in filters, and not in from_date arg.
we need to catch it, and format date as magento api needs

### DIFF
--- a/magentoerpconnect/product.py
+++ b/magentoerpconnect/product.py
@@ -198,8 +198,13 @@ class ProductProductAdapter(GenericAdapter):
         """
         if filters is None:
             filters = {}
+        str_from_date = ''
         if from_date is not None:
             str_from_date = from_date.strftime('%Y/%m/%d %H:%M:%S')
+        if 'from_date' in filters:
+            str_from_date = filters['from_date'].strftime('%Y/%m/%d %H:%M:%S')
+            filters.pop('from_date')
+        if str_from_date:
             filters['updated_at'] = {'from': str_from_date}
         # TODO add a search entry point on the Magento API
         return [int(row['product_id']) for row


guewen:
Where is the code that put 'from_date' in the filters? Maybe it would be simpler to put an assert to prevent having this key in filters?

rdeheele:
via button import_product_product..

<pre>
     def import_product_product(self, cr, uid, ids, context=None):
        self._import_from_date(cr, uid, ids, 'magento.product.product',
                               'import_products_from_date', context=context)
        return True
</pre>

that calls _import_from_date
in connector-magento / magentoerpconnect / magento_model.py

<pre>
    def _import_from_date(self, cr, uid, ids, model, from_date_field,
                          context=None):
...
        for backend in self.browse(cr, uid, ids, context=context):
            from_date = getattr(backend, from_date_field)
            if from_date:
                from_date = datetime.strptime(from_date,
                                              DEFAULT_SERVER_DATETIME_FORMAT)
            else:
                from_date = None
            import_batch.delay(session, model,
                               backend.id, filters={'from_date': from_date})
</pre>

I don't see the case where we can have from_date as a solitary argument in search function.
The methods def import_batch, then def run in import_synchronizer takes only one filters argument.

guewen:
The code you shows is import_batch, not the method search.

```
        import_batch.delay(session, model,
                           backend.id, filters={'from_date': from_date})
```

`from_date` is popped out of the filters of the import_batch to be passed to the method `search` of the adapter

https://github.com/OCA/connector-magento/blob/7.0/magentoerpconnect/product.py#L254

There is a distinction between the filters applied on the batch import - as seen by openerp - and the filters that we pass to magento, that are not necessarily the same than the filters internal to the batch import.
The `from_date` argument exists because it needs a processing before being passed as a Magento filter (here the format conversion). The values in `filters` are raw Magento filters. IMO if from_date is passed directly in filters, it should be the format expected by Magento.

rdeheele:
ok, it seems that the problem that I met comes from a break in inheritance mecanism.
when the job "Prepare a batch import of records from Magento" is executed:
it passed in def run in BatchImportSynchronizer(ImportSynchronizer) class, and not in ProductBatchImport(DelayedBatchImport) class. for the moment, I don't see why...

rdeheele:
mmh, certainly comes from magento decorator...

guewen:
@rdeheele do you have custom code?

What is the `_model_name` of `BatchImportSynchronizer`?

Issue #34, opened 2014-09-10
magentoerpconnect_order_comment: nltk.clean_html() raises a NotImplementedError


guewen:
magentoerpconnect_order_comment uses `nltk.clean_html(comment)` which does no longer work. It has been removed from the lib in version 3.0: https://github.com/nltk/nltk/commit/39a303e5ddc4cdb1a0b00a3be426239b1c24c8bb

From Travis:

```
2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: ERROR: test_20_export_sale_comment (openerp.addons.magentoerpconnect_order_comment.tests.test_synchronization.TestMagentoSaleCommentExport)

2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: ` Test export of sale order comment

2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: Traceback (most recent call last):

2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: ` File "/home/travis/build/OCA/connector-magento/magentoerpconnect_order_comment/tests/test_synchronization.py", line 154, in test_20_export_sale_comment

2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: ` export_record(self.session, 'magento.sale.comment', mag_comment_id)

2014-09-10 10:23:45,300 23456 TEST openerp_test openerp.modules.module: ` File "/home/travis/build/OCA/connector-magento/magentoerpconnect/unit/export_synchronizer.py", line 389, in export_record

2014-09-10 10:23:45,301 23456 TEST openerp_test openerp.modules.module: ` return exporter.run(binding_id, fields=fields)

2014-09-10 10:23:45,301 23456 TEST openerp_test openerp.modules.module: ` File "/home/travis/build/OCA/connector-magento/magentoerpconnect/unit/export_synchronizer.py", line 124, in run

2014-09-10 10:23:45,301 23456 TEST openerp_test openerp.modules.module: ` result = self._run(*args, **kwargs)

2014-09-10 10:23:45,301 23456 TEST openerp_test openerp.modules.module: ` File "/home/travis/build/OCA/connector-magento/magentoerpconnect/unit/export_synchronizer.py", line 375, in _run

2014-09-10 10:23:45,301 23456 TEST openerp_test openerp.modules.module: ` record = self._create_data(map_record, fields=fields)

[...]

PR #39, opened 2014-09-24
Prevent to create duplicate bindings on invoices

guewen:
The correction made at becc1bf2 was not working because it was
doing a 'continue' on the inner loop but not on the outer loop.


guewen:
Travis fails because of https://github.com/OCA/connector-magento/issues/34

nbessi:
:+1:

guewen:
Rebased so #34 will no longer make travis fail.

OSguard:
:+1:

PR #41, opened 2014-10-03
[IMP] improve debug log. Add logger before calling Magento API.

sebastienbeau:
Indeed if not we do not know want we look in the log what OpenERP is waiting. Add the processing time of the call, in order to have some metric about how slow is Magento ;)

### DIFF
--- a/magentoerpconnect/unit/backend_adapter.py
+++ b/magentoerpconnect/unit/backend_adapter.py
@@ -27,7 +27,7 @@ import magento as magentolib
 from openerp.addons.connector.unit.backend_adapter import CRUDAdapter
 from openerp.addons.connector.exception import (NetworkRetryableError,
                                                 RetryableJobError)
-
+from datetime import datetime
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


guewen:
While you are modifying this logger, a nice thing could be something like:

```
try:
    result = api.call(method, arguments)
except:
    _logger.debug("api.call(%s, %s) failed", method, arguments)
    raise
else:
    _logger.debug("api.call(%s, %s) returned %s in %s seconds",
    method, arguments, result,
    (datetime.now() - start).seconds) 
```

What do you think?

sebastienbeau:
Hi Guewen, My aim was not to catch error, but to give a better feedback for the developper.
If you are developing and look at the log you will see the log of the call after the execution of the call. For a newbie, it can be hard for him to understand what happen. Indeed you will see that OpenERP (with the log) start the job, process the mapping.... and them you will wait a long time and see the log of the call to Magento. I prefer to see the log before the call (for now it's after). Indeed if you see the log before the call, just be looking to the long you know that a call is in progress and you will have to wait for Magento. What do you think?

sebastienbeau:
Regarding the catch of the error, It's really a good idea ! I will add the change

florian-dacosta:
Seems good to me! thanks :+1:

bguillot:
Looks good, thanks :+1:

guewen:
:+1:

PR #42, opened 2014-10-05
Refactor code in order to be compatible with fiscal position rule module

sebastienbeau:
Partner and address must be set correctly before calling the onchange method.
With this refactor and this other PR https://github.com/OCA/connector-ecommerce/pull/7, the fiscal position are set correctly on sale order when importing a sale order

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -695,7 +695,7 @@ class SaleOrderImport(MagentoImportSynchronizer):
         self.partner_invoice_id = billing_id
         self.partner_shipping_id = shipping_id or billing_id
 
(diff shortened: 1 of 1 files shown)


guewen:
I was sure it was already working, anyway the code is better know I think, using the options of the mapper.

Thanks

:+1:

sebastienbeau:
A case that was not working was the case where you create a customer with a Suisse address as main address and send the delivery to France.
In this case the address used for setting the fiscal position will be the suisse one and not the french one (In France fiscal position are based on delivery address).

bguillot:
Thanks

:+1:

florian-dacosta:
:+1:

PR #47, opened 2014-10-06
7.0 connector magento rma

sebastienbeau:
Add module for synchronizing RMA
We have a basic module on magento side https://code.launchpad.net/~magentoerpconnect-core-editors/magentoerpconnect/magento-rma-module

Question : maybe we should move all magento-module in github too? (under OCA?)


magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+
+    def search_read(self, filters=None, from_date=None):
+        """ Search records according to some criterias
+        and returns their information
+        """
+        backend = self.session.browse('magento.backend', self.backend_record.id)
+        if filters is None:
+            filters = {}
+        if from_date is not None:
+            filters = from_date.strftime('%Y-%m-%d %H:%M:%S')
```
guewen: I don't think filter should be a date as string but a {}.

magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+        elif backend:
+            filters = backend.import_claims_from_date
```
guewen, line 404: Same as before, but furthermore the date is not in the magento datetime format

magentoerpconnect_rma/claim.py
guewen: Also, the magento datetime format is available as a constant in magentoerpconnect

magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+        return self._call('%s.list' % self._magento_model,
+                          [filters, 1] if filters else [{}, 1])
+
+    def read(self, magento_record, attributes=None):
+        """ Returns the information of a record
+
+        :rtype: dict
+        """
+        return magento_record
```
guewen: ?

magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+        """ Search records according to some criterias
+        and returns their information
+        """
+        backend = self.session.browse('magento.backend', self.backend_record.id)
+        if filters is None:
+            filters = {}
+        if from_date is not None:
+            filters = from_date.strftime('%Y-%m-%d %H:%M:%S')
+        elif backend:
+            filters = backend.import_claims_from_date
```
guewen, line 447: Same comments as for ClaimCommentAdapter

magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+        return self._call('%s.list' % self._magento_model,
+                          [filters, 1] if filters else [{}, 1])
+
+    def read(self, magento_record, attributes=None):
+        """ Returns the information of a record
+
+        :rtype: dict
+        """
+        return magento_record
```
guewen: ?

magentoerpconnect_rma/claim.py
```diff
@@ -0,0 +1,1055 @@
...
+
+    def run(self, filters=None):
+        """ Run the synchronization """
+        from_date = filters.pop('from_date', None)
+        record_ids = self.backend_adapter.search(filters, from_date)
+        formated_date = ''
+        if from_date:
+            formated_date = from_date.strftime('%Y-%m-%d %H:%M:%S')
+        _logger.info('search for magento claims %s returned %s',
+                     formated_date, record_ids)
```
guewen: `formated_date` is used only for the logger? the code is duplicated with the one in `search`, why not log into `search`?

Issue #49, opened 2014-10-15
[7.0][Multi company][Sale order import] Wrong taxes on order lines 


kenvac:
Hi Team,

I was testing sale order import with multi company with different taxes and found a bug.

Steps to reproduce:
- Create 2 companies VIZ UK, US (Main company)
- Setup tax T1 0% (US) and T2 20% (US)
- Create a category C. Assign income Account and other relavant property accounts on category. Repeat this step for other company as well.
- Open income account (US) and set Default Taxes - T1 0% 
- Open income account (UK) and set Default Taxes - T2 0% 
- Create a product A under category C
- Setup two websites on magento UK and US with store views
- Create a sale order on UK webstore
- Import SO
- Result, the sale order uses US takes on UK sale order.

Problem:
- Shop not passed in context on product onchange

Solution:
set shop in session context.

guewen:
Hi @kenvac,

Using the Magento connector in multi company has probably several bugs such as this one because nobody implemented it yet. It has been discussed though. I think that your bug comes from the fact that the connector uses the 'admin' user who has all the access rights. Thereby, the record rules based on the company of the user are not applied. I added some advices on one of your PR:  https://github.com/OCA/connector-magento/pull/52#issuecomment-59723467

Let me know if you need help.

PR #50, opened 2014-10-15
[FIX] Wrong taxes on order lines

kenvac:
Fix for issue #49


guewen:
I think that implementing this would help https://github.com/OCA/connector-magento/pull/52#issuecomment-59723467.
Otherwise, same comment than: https://github.com/OCA/connector-magento/pull/52#discussion_r19076756

Thanks for your proposal.

pedrobaeza:
It seems that this is not going to be attended, so I close.

Issue #51, opened 2014-10-15
[7.0][multicompany][Payment methods] Payment methods not searched with company


kenvac:
Hi, 

I've a scenario were there are three websites on magento using same payment method. 
In this case, the wrong payment method is set on sale order.

PR #52, opened 2014-10-15
7.0 search payment methods with company

kenvac:
Fix for issue #51

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -905,6 +905,7 @@ class SaleOrderImportMapper(ImportMapper):
                                           'non existing storeview')
         storeview = self.session.browse('magento.storeview', storeview_id)
         shop_id = storeview.store_id.openerp_id.id
+        self.session.context['shop'] = shop_id
         return {'shop_id': shop_id}
 
     @mapping
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -926,10 +926,10 @@ class SaleOrderImportMapper(ImportMapper):
             assert storeview_id is not None, 'cannot import sale orders Payment from non existing storeview'
             storeview = self.session.browse('magento.storeview', storeview_id)
             company_id = storeview.store_id.openerp_id.company_id or False
-            self.session.context['company_id'] = company_id
-            self.session.context['force_company'] = company_id
             if company_id:
                 company_id = company_id.id
+                self.session.context['company_id'] = company_id
+                self.session.context['force_company'] = company_id
         method_ids = self.session.search('payment.method',
                                 [['name', '=', record['payment']['method']],
                                 ['company_id', '=', company_id]])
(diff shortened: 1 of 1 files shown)


magentoerpconnect/sale.py
```diff
@@ -905,6 +905,7 @@ def store_id(self, record):
                                           'non existing storeview')
         storeview = self.session.browse('magento.storeview', storeview_id)
         shop_id = storeview.store_id.openerp_id.id
+        self.session.context['shop'] = shop_id
```
guewen, line 908: The context should not be mutated directly in the session, it should be changed with `with self.session.change_context(...):`. It should not be modifier in the Mapper since the Mapper should only take data in and return data out, and not use a side effect on the state of the Importer. The Importer has the responsibility to change the context for example before calling the creation or update of the record.

guewen:
The multi-company is currently not well supported and this fix is aiming to a better support, so thanks for you proposal. However, this is not the approach I would take. Here is mine:

Each website or store view should be linked to a company, and each company should have a new field `connector_user_id`. This field is a relation to a user who is himself assigned to one company. Seems a bit strange but there is a reason: when a website / store view has a company and this company has a `connector_user_id`, the connector will be able to switch to this user, letting OpenERP apply record rules and properties according to this user. The domain should not be modified in the mapper because the record rule will automatically restrict on the allowed payment methods.

Here is an example of implementation: https://gist.github.com/guewen/3528fe4be6bf527ed9d9 and a thread discussing that: https://www.mail-archive.com/openerp-connector-community@lists.launchpad.net/msg00091.html

The big benefit of this approach is that we don't need to change every domain in all the mappings to ensure they filter on the company. This is also the only way to be sure that the property fields are set correctly and the internal OpenERP code (function fields, ...) use the correct company.

magentoerpconnect/sale.py
kenvac, line 908: I'll change that one. Thanks.

kenvac:
guewen, Thanks for spending your valuable time and putting comment here. 
I got your point of setting up new users per company. But, how do I configure them to use with per storeview/store/website?  It might be easy, sorry for being so naive.

guewen:
When importing a sales order, it is linked to a shop which has a `company_id` field, so you will assign the right company to each shop.

In the Magento importer, it will give something like that:

```
@magento
class SaleOrderImport(MagentoImportSynchronizer):
    _model_name = ['magento.sale.order']

    def _import(self, binding_id):
        binder = self.get_binder_for_model('magento.storeview')
        storeview_id = binder.to_openerp(record['store_id'])
        assert storeview_id is not None, ('cannot import sale orders from '
            'non existing storeview')
        storeview = self.session.browse('magento.storeview', storeview_id)
        shop = storeview.store_id
        user = shop.company_id.connector_user_id
        if user:
            with self.session.change_user(user.id):
                super(SaleOrderImport, self)._import(binding_id)
        else:
             super(SaleOrderImport, self)._import(binding_id)
```

mistotebe:
How does a new user per company play with Odoo pricing? Do you know if such users would be considered active and charged for under the support contract? If so, this might be an issue for some customers who have several companies but only a few users per company (and none of those user can be reused for magento sync).

guewen:
@mistotebe only users who have logged in at least 1 time should be charged.

pedrobaeza:
Is this going to be continued?

mistotebe:
Unlikely, there is no v7 multi-company with Magento integration that we need it for anymore.

PR #53, opened 2014-10-16
Add modular domain to update stk in magento

bealdav:
end exclude service product


bealdav:
should be on v7 first, sorry

PR #57, opened 2014-10-25
[IMP] skip the export if the binding do not exist anymore

sebastienbeau:
Sometime the user can create by error a binding and delete it just after. If the job have been not processed, the job will failed. We should skip it as there is nothing to export anymore

### DIFF
--- a/magentoerpconnect/unit/export_synchronizer.py
+++ b/magentoerpconnect/unit/export_synchronizer.py
@@ -389,6 +389,8 @@ class MagentoExporter(MagentoBaseExporter):
 @related_action(action=unwrap_binding)
 def export_record(session, model_name, binding_id, fields=None):
     """ Export a record on Magento """
+    if not session.search(model_name, [['id', '=', binding_id]]):
+        return "The binding do not exist anymore, skip it"
     record = session.browse(model_name, binding_id)
     env = get_environment(session, model_name, record.backend_id.id)
     exporter = env.get_connector_unit(MagentoExporter)


guewen:
Not sure about that. What about inactive records? Should they be skipped as well (this is what will happen with your change)?

guewen:
@sebastienbeau do you have an idea for my question?

sebastienbeau:
Sorry
For managing active/inactive, here is my idea.
I am planning to add a active field on the binding. If I inactive a record I will automatically inactive the binding. 
Inativating a binding will call the unlink method on Magento. Indeed when you have a big rotation of product in your catalog (my case is around 30K€ product per year) you need to clear your magento catalog. In Odoo we always keep the product and the binding but in inactivated state.
In case that we reactive the product, it will create a new binding (the product will have one activated binding and one inactivated with the old id). With this solution we keep an history of old id and so we can still reimport an old order if necessary.

So my conclusion, is I will never export an inactivating record as this record will be deleted on Magento
What do you think?
Thanks

guewen:
You speak about products but your changes impacts all types of records.

guewen:
> In case that we reactive the product, it will create a new binding (the product will have one activated binding and one inactivated with the old id)

It won't work as long as the SKU must be unique

sebastienbeau:
I think you are right, we should only add this for some special case.

> It won't work as long as the SKU must be unique
> I think my explication was not clear.

Let's see an exemple.
I have the product A activated on Odoo.
When I desactivate the binding in Odoo, it will remove it on Magento (the product is deleted so the sku do not exist anymore)

If I create a new binding (because I want to sell again the product). The product will be exported in Magento (with the same sku as before, as the product was deleted before there is not sku conflict)

At the end I have
- one inactivated binding (which store the old magento id)
- one activated binding (which store the new magento id)

Hope that my explication are more clear

guewen:
> It won't work as long as the SKU must be unique
> I think my explication was not clear.

I meant "unique in Odoo".

sebastienbeau:
So my explication was not clear ;) . In Odoo we wil have only 1 product with 2 magento binding (one inactivated; one activated) and at the end 1 product in Magento.

guewen:
Understood. BTW on the same topic, we may want to add a constraint that check if a product has no more than 1 active binding for the same backend.

sebastienbeau:
Yes I agree :)

PR #68, opened 2014-11-26
don't match res.partner.title on substrings

tremlin:
I would like to suggest to use the =ilike operator for title matching because at least in our installations we want to distinguish between "Herr" and "Herr Dr.", for instance.


guewen:
:+1:  but needs a rebase it seems

Thanks

tremlin:
I rebased my patch.

OSguard:
:+1:
builds fails to pylint

pedrobaeza:
:+1:

guewen:
I restarted the build and it is now green

PR #70, opened 2014-12-02
Catch cases where record['discount_amount'] = None

mdietrichc2c:
Magento sometimes sends sales_order.items with the attributes 'discount_amount' set to None; this caused the float() function to fail with the error "TypeError: float() argument must be a string or a number".

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -992,10 +992,14 @@ class SaleOrderLineImportMapper(ImportMapper):
 
     @mapping
     def discount_amount(self, record):
-        discount_value = float(record.get('discount_amount', 0))
-        if self.backend_record.catalog_price_tax_included:
+        discount_value = 0.0
+        row_total = 0.0
+        if record.get('discount_amount'):
+            discount_value = float(record.get('discount_amount', 0))
+        if self.backend_record.catalog_price_tax_included and \
+           record.get('row_total_incl_tax'):
             row_total = float(record.get('row_total_incl_tax', 0))
-        else:
+        elif record.get('row_total'):
             row_total = float(record.get('row_total', 0))
         discount = 0
         if discount_value > 0 and row_total > 0:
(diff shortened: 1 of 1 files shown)


magentoerpconnect/sale.py
```diff
@@ -992,10 +992,14 @@ class SaleOrderLineImportMapper(ImportMapper):
...
     @mapping
     def discount_amount(self, record):
-        discount_value = float(record.get('discount_amount', 0))
-        if self.backend_record.catalog_price_tax_included:
+        discount_value = 0.0
+        row_total = 0.0
+        if record.get('discount_amount'):
+            discount_value = float(record.get('discount_amount', 0))
+        if self.backend_record.catalog_price_tax_included and \
+           record.get('row_total_incl_tax'):
```
pedrobaeza: Bad indentation

guewen: Also do not put a `\` but use parentheses

magentoerpconnect/sale.py
```diff
@@ -992,10 +992,14 @@ class SaleOrderLineImportMapper(ImportMapper):
 
     @mapping
     def discount_amount(self, record):
-        discount_value = float(record.get('discount_amount', 0))
-        if self.backend_record.catalog_price_tax_included:
+        discount_value = 0.0
+        row_total = 0.0
+        if record.get('discount_amount'):
+            discount_value = float(record.get('discount_amount', 0))
```
guewen: `discount_value = float(record.get('discount_amount') or 0)` is simpler

guewen:
I propose to leave the existing code but just use this idiom `value = float(record.get('field_amount') or 0)` instead of `value = float(record.get('field_amount', 0))` (also in the other amount fields).

mdietrichc2c:
@guewen , @pedrobaeza : yes, it's cleaner like this. I fixed it.

guewen:
Thanks!

:+1:

pedrobaeza:
:+1:

Issue #72, opened 2014-12-11
Separate tax settings per storeview


mistotebe:
It appears that Magento can have different settings for each storeview with regards whether tax is included in the price or not, which is a situation the connector cannot handle at the moment. Not sure if there is a way to find out what the setting is from magento through the API.

Proposed solution: add a checkbox (and a migration script to move the backend setting there) or a tri-state field (included/excluded/inherit from backend) on the storeview and update the code in SO import to check that instead.

What do you think?

mistotebe:
If I go with the checkbox on storeview+migration script, what is the versioning policy of the module? Do changes like this bump the minor or major version (aka 2.4.3 or 2.5.0)? I'm going to guess 2.4.3, but making sure.

guewen:
We can go with 2.5.0 as schema will change. Before the migration to GitHub, I used to merge all the things that change the schema in one "big" release (so 2.4.x to 2.5.0 for instance) and fixes that just needed to restart the server in a minor release (2.4.2 to 2.4.3). But now that the project is on GitHub with the merge one-click-button and that any committer within the OCA can push the button, it is more difficult to have true releases as merges are spread over time.

Regarding the change you propose, I am not the one with the most knowledge on tax include/exclude, but for what I know it makes sense. 

ping @rdeheele, can you confirm?

> Not sure if there is a way to find out what the setting is from magento through the API.

Worth to check but I would be surprised if it was possible ;-)

mistotebe:
Talked to a person on our team who knows some magento and looks like getting that setting from magento is not (easily) possible.

guewen:
So it has to be configured manually, that's not the first nor the last thing we won't be able to sync automatically with magento :-(

rdeheele:
yes, this setting about tax inclusion seems not available through the API.
If the tax inclusion can vary according the storeview in Magento, it makes sense to be symetric in Odoo.

PR #74, opened 2014-12-12
Tax inclusion settings per storeview

mistotebe:
Closes issue #72.

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -599,20 +599,23 @@ class SaleOrderImport(MagentoImportSynchronizer):
                 SaleOrderMoveComment)
             move_comment.move(binding)
 
+    def _get_storeview(self, record):
+        """ Return the tax inclusion setting for the appropriate storeview """
+        storeview_binder = self.get_binder_for_model('magento.storeview')
+        # we find storeview_id in store_id!
+        # (http://www.magentocommerce.com/bug-tracking/issue?issue=15886)
+        storeview_id = storeview_binder.to_openerp(record['store_id'])
+        storeview = self.session.browse('magento.storeview', storeview_id)
+        return storeview
+
     def _get_magento_data(self):
         """ Return the raw Magento data for ``self.magento_id`` """
         record = super(SaleOrderImport, self)._get_magento_data()
         # sometimes we don't have website_id...
         # we fix the record!
         if not record.get('website_id'):
+            storeview = self._get_storeview(record)
             # deduce it from the storeview
-            storeview_binder = self.get_binder_for_model('magento.storeview')
-            # we find storeview_id in store_id!
-            # (http://www.magentocommerce.com/bug-tracking/issue?issue=15886)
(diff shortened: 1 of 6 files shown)


magentoerpconnect/sale.py
```diff
@@ -780,7 +787,7 @@ def _create_data(self, map_record, **kwargs):
             **kwargs)
 
     def _update_data(self, map_record, **kwargs):
-        tax_include = self.backend_record.catalog_price_tax_included
+        tax_include = self._get_price_included(map_record)
         self._check_special_fields()
         return super(SaleOrderImport, self)._update_data(
             map_record, tax_include=tax_include,
```
guewen: Can I propose a small refactoring for this PR? I won't blame you if you can't afford to change that, your work is well done and my suggestion goes further than your initial goal.

So my feeling is that having the Storeview record in the "options" of the mapper would be interesting because several mapping methods may need it. Currently, each mapping method have to do the same code:

```
storeview_binder = self.get_binder_for_model('magento.storeview')
storeview_id = storeview_binder.to_openerp(record.source['store_id'])
storeview = self.session.browse('magento.storeview', storeview_id)
```

My suggestion is to replace `_get_price_included` by a `_get_storeview` method that returns the storeview record and pass it in `super(SaleOrderImport, self)._update_data()` and `super(SaleOrderImport, self)._create_data()`. Then, the mapper could use `self.options.storeview.catalog_price_tax_included` instead of `self.options.tax_include`. (But maybe we want to keep the same value as well in `self.options.tax_include` just for backward compatibility).

guewen:
Thanks @mistotebe! Well done.

Let me know what you think about my suggestion in the diff lines.

mistotebe:
I guess runbot is broken? I have not seen a successful build in there at all (those that are not failed seem they have not actually run at all or have never finished), also I cannot seem to access the build logs to check why it thinks it should not pass.

guewen:
@mistotebe yes, don't mind with runbot, it is not working correctly yet. By cons, Travis did not run, probably due to a conflict.

mistotebe:
Do you mind if I rebase then?

mistotebe:
(and squash, I guess)

guewen:
Excellent! :+1: 

BTW you can add your name in https://github.com/OCA/connector-magento/blob/7.0/magentoerpconnect/AUTHORS

Thanks

mistotebe:
Eventually, or I'll let you do that after this gets merged. No point in adding it to the merge request at this point.

guewen:
Ok. I just need one other person to approve in order to merge it (OCA rules).

Ping @OSguard @sebastienbeau @rdeheele

rdeheele:
:+1: 
thanks @mistotebe , good work!

guewen:
@mistotebe I'm going to merge right now. Do you want me to add a line in AUTHORS? Do you want me to put your real name (that I don't have) or mistotebe?

mistotebe:
@guewen, sure, why not. Looking at the file, the line should probably read "\* Ondřej Kuzník at credativ"

guewen:
Alright

PR #77, opened 2015-01-19
Fix unterminated string in de.po

yvaucher:
Fix error raised by rosetta import on launchpad:

```
Line 126: String not terminated
```

And found a second one


pedrobaeza:
:+1:

guewen:
:+1:

PR #85, opened 2015-02-19
Migration of magentoerpconnect

guewen:
# Phase:

**Needing reviews!**
# How to build yourself

I use this configuration to have all the required pull requests merged https://github.com/guewen/odoo-connector-magento-buildout/tree/8.0
# What is in there
## Changes
- Migrated on the new Odoo API and the new Connector API
- Extended the test coverage
- Show the failed calls in the logger with the ERROR level
- New Docker image for Magento, mentionned in the documentation
- Allow to choose a default sales team for sales orders on the storeviews
- Lot of improvements/cleaning of the code
- The binders can now return records (new `browse` argument)
- The Importer now pass the binding report instead of the binding_id
  internally
- Remove sale.shop
## More details on the API changes

First have a look on OCA/connector#47 for details on the Connector changes.
### Binders
- A new argument allows to return records instead of ids in `to_openerp()`. `to_backend()` and `bind()` both accept ids and records as input.
### Importer
- `get_binding_id()` has been renamed to `get_binding` and returns a binding instead of an ID. It also means that internally, the methods should now work with the record and not with the ID.
- The base synchronizers have been renamed from `MagentoImportSynchronizer` to `MagentoImporter` and `MagentoExportSynchronizer` to `MagentoExporter` 
- All the specialized importers/exporters have been renamed to `SomethingImporter`, instead of `SomethingImport`

PR #86, opened 2015-02-24
Use Mapper's options to pass the parent partner to the address Mapper

guewen:
The parent_id field in 'record' is a Magento ID!

Backported from #85

Correct commit 5fad5540
Credits @jssuzanne for the analysis of the bug


OSguard:
:+1:

StefanRijnhart:
:+1:

PR #90, opened 2015-03-11
Add the original authors in the magento connector extension...

guewen:
...Modules. Since we can now add several authors.

And set module magentoerpconnect_catalog as uninstallable

Because it does nothing and appears on apps.odoo.com as a valid module
otherwise.


pedrobaeza:
:+1: with the same comment as with e-commerce.

jgrandguillaume:
:+1:

guewen:
@pedrobaeza I removed the generic authors

yvaucher:
:+1:

PR #110, opened 2015-07-03
Setup transifex

guewen:


pedrobaeza:
Change coveralls

pedrobaeza:
Runbot is failing because magento lib is not installed. As @gurneyalex is on holidays, someone with SSH access to the runbot host must install it

guewen:
@pedrobaeza I was about to change it, you was faster than me ;-)

Done the changes

pedrobaeza:
:+1:

max3903:
:+1: 

I just installed magento lib on runbots

guewen:
Thanks @max3903 
I restarted the runbot build

guewen:
Runbot says `ImportError: No module named connector_ecommerce`
Did I miss something with the dependencies?

pedrobaeza:
connector-ecommerce is put as dependency, so it should work. I have restarted again to see.

pedrobaeza:
The error is because you are using `from openerp.addons.connector_ecommerce.unit.sale_order_onchange import...` expression, which maybe on the runbot architecture cannot be resolved due to python paths and so on, but I don't know enough the internals of runbot to know it.

guewen:
> The error is because you are using from openerp.addons.connector_ecommerce.unit.sale_order_onchange import... expression, which maybe on the runbot architecture cannot be resolved due to python paths and so on, but I don't know enough the internals of runbot to know it.

This import is correct though.

PR #115, opened 2015-07-03
Wrong attribute name for the shipping line builder

### LINKED ISSUE #112: While import sale order, correct product is not set inside shipping line according to carrier method.
While import sale order from Magento to Odoo, always default product "[SHIP]" is set inside shipping line of that order. Even though we have specified product inside carrier method of that order.

System is not going to getting correct product from selected carrier method of that order instead of system always taking default service product which have external id "connector_ecommerce.product_product_shipping". 

We have been tested with Odoo V8. Also getting it inside OpenERP V7.

guewen:
Fixes #112
Backport from #114

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -890,7 +890,7 @@ class SaleOrderImportMapper(ImportMapper):
         if values.get('carrier_id'):
             carrier = self.session.browse('delivery.carrier',
                                           values['carrier_id'])
-            line_builder.product_id = carrier.product_id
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@guewen is this still needed?

guewen:
Yes it is

pedrobaeza:
Do you know why Travis is red?

guewen:
I restarted it this afternoon as it didn't run since a long time... Well, test_pylint is failing and `test_server.py` does not even start. I'll try to check

PR #95, opened 2015-04-09
7.0 bundle import base

sebastienbeau:
Migrate this merge proposal from lp : https://code.launchpad.net/~akretion-team/openerp-connector-magento/7.0-next-release-bundle-import-base/+merge/223211

This Pull Request depends of pull requests:
- [ ] OCA/stock-logistics-workflow#101 
- [ ] OCA/account-invoicing#69

PR #117, opened 2015-07-09
7.0 bundle import child

OSguard:
First Version for importing bundle as child


magentoerpconnect_bundle_import_childs/backend.py
```diff
@@ -0,0 +1,25 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+import openerp.addons.connector.backend as backend
+import openerp.addons.magentoerpconnect.backend as magento_backend
+
+magento1700_bundle_import_childs = backend.Backend(parent=magento_backend.magento1700,
+                                    version='1.7-bundle-import-childs')
```
guewen, line 25: You shouldn't create a new backend version for a new feature (otherwise you will have to choose which version you want and it will allow only 1 feature at a time ;-))
Instead, you should rather use the `@magento(replacing=OtherConnectorUnit)` in order to modify the base behavior. When you do that, you must isolate only the behavior you must to replace in its own ConnectorUnit (e.g. do not replace ProductMapper but an isolated ProductBundleMapper). In fact there is already a pending basis for this which isolates the pieces: https://github.com/OCA/connector-magento/pull/95

chafique-delli:
@OSguard: Hi, we have developed a basic module for import bundles (https://github.com/OCA/connector-magento/pull/95)  that can serve you.
Thanks,

pedrobaeza:
CIs statuses are red and very old, so I close it. If there's still interest in this, feel free to open a new PR.

PR #128, opened 2015-08-12
remove deprecated call to install_in_connector

gurneyalex:
the sole presence of connector.py is all that is needed for the addon to be
considered.

See https://github.com/OCA/connector/pull/74

### DIFF
--- a/__unported__/magentoerpconnect_export_partner/connector.py
+++ b/__unported__/magentoerpconnect_export_partner/connector.py
@@ -19,7 +19,4 @@
 #
 ##############################################################################
 
-from openerp.addons.connector.connector import install_in_connector
-
-
-install_in_connector()
+pass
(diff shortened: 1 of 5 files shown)


__unported__/magentoerpconnect_export_partner/connector.py
```diff
@@ -19,7 +19,4 @@
 #
 ##############################################################################
 
-from openerp.addons.connector.connector import install_in_connector
-
-
-install_in_connector()
+pass
```
pedrobaeza, line 22: Isn't enough to put an empty py?

nbessi:
:+1:

__unported__/magentoerpconnect_export_partner/connector.py
gurneyalex, line 22: it probably is.

pedrobaeza:
:+1: in any case, because the remark is not very important

guewen:
The presence of the file `connector.py` is in fact not needed.

guewen:
Please see my takeover: #130
I deleted the empty `connector.py` files.

PR #130, opened 2015-08-17
Remove deprecated calls to install_in_connector

guewen:
See https://github.com/OCA/connector/pull/74

Takeover of #128 because @gurneyalex is away for one week.


pedrobaeza:
:+1:

bguillot:
:+1:

PR #131, opened 2015-08-17
Be more specific in the notice

### LINKED ISSUE #126: Dependency up to date ?
In the documentation it is written : 

```

Important

Please check if you have installed Magento on PHP with a 5.4.x* version.
Magento is not compatible with this version and would prevent the API to behave normally. 
In that case, you must retrograde to PHP 5.3.x or apply the patch provided by Magento (see http://magento.com/resources/system-requirements)
```

http://magento.com/resources/system-requirements tells something different. Is the doc obsolete or is there an other subtlety?

guewen:
Because it does not apply for Magento 1.9

Closes #126

### DIFF
--- a/magentoerpconnect/doc/guides/installation_guide.rst
+++ b/magentoerpconnect/doc/guides/installation_guide.rst
@@ -83,8 +83,9 @@ In order to install it:
    `magento_root/var/cache`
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

Issue #129, opened 2015-08-13
Wrap synchronyze_metadata to avoid exceptions


danimaribeiro:
What about wrap this with try except and give a nice message?
![syncrhonize](https://cloud.githubusercontent.com/assets/854781/9260109/fdf501f2-41db-11e5-99d1-6b736de47c05.PNG)

To avoid errors like this: 
https://github.com/OCA/connector-magento/issues/122

guewen:
Good idea, it would be nice indeed.

danimaribeiro:
Ok, a PR is coming :)

PR #133, opened 2015-08-20
[FIX] #129 - Wrap synchonize metadata method and give the user a warning

danimaribeiro:


StefanRijnhart:
:+1:

magentoerpconnect/magento_model.py
```diff
@@ -186,17 +187,23 @@ def check_magento_structure(self):
...
-                          'magento.store',
-                          'magento.storeview'):
-                # import directly, do not delay because this
-                # is a fast operation, a direct return is fine
-                # and it is simpler to import them sequentially
-                import_batch(session, model, backend.id)
-        return True
+        try:
+            session = ConnectorSession(self.env.cr, self.env.uid,
+                                       context=self.env.context)
```
guewen, line 192: Could you take the opportunity of this diff to improve that line to `ConnectorSession.from_env(self.env)`?

magentoerpconnect/magento_model.py
```diff
@@ -186,17 +187,23 @@ def check_magento_structure(self):
...
+                    # import directly, do not delay because this
+                    # is a fast operation, a direct return is fine
+                    # and it is simpler to import them sequentially
+                    import_batch(session, model, backend.id)
+            return True
+        except Exception as e:
+            _logger.error(e.message, exc_info=True)
+            raise Warning(
+                _(u"Check your configuration, we can't get the data.\
+                Here is the error:\n%s") % str(e).decode('utf-8', 'ignore'))
```
guewen, line 206: Could you write it in the following manner:

``` python
raise Warning(
    _(u"Check your configuration, we can't get the data. "
      u"Here is the error:\n%s") % str(e).decode('utf-8', 'ignore'))
```

So we don't get blanks in the string?

guewen:
Thanks for the contrib!

StefanRijnhart:
Superseded by https://github.com/OCA/connector-magento/pull/231

PR #151, opened 2015-09-24
Check street before split()

mdietrichc2c:
Small PR to add a check before mapping the partner's street' attribute. This avoids an exception when `.split('\n')` is called on an empty value.

### DIFF
--- a/magentoerpconnect/partner.py
+++ b/magentoerpconnect/partner.py
@@ -514,6 +514,8 @@ class BaseAddressImportMapper(ImportMapper):
 
     @mapping
     def street(self, record):
+        if not record.get('street'):
+            return
         value = record['street']
         lines = [line.strip() for line in value.split('\n') if line.strip()]
         if len(lines) == 1:


magentoerpconnect/partner.py
```diff
@@ -514,6 +514,8 @@ def country(self, record):
 
     @mapping
     def street(self, record):
+        if not record.get('street'):
+            return
```
pedrobaeza, line 518: You should return `{}`

guewen, line 518: No, `return` is fine

guewen:
:+1:

pedrobaeza:
OK then, although return type is not consistent.

:+1:

guewen:
> OK then, although return type is not consistent.

As far as I agree that a function should never return 2 different types (like a string or a list), returning one type or `None` (like `re.search` does) is reasonable. The returned type is `None` which is perfectly handled by the Mapper (by the way, the return of the function is only handled by the internals of the Mapper), it makes the mapping interface smoother to use (you have to return a dict only if you want to change a value) and as a positive side-effect, prevent unnecessary dicts allocations .

Not the object of this PR anyway.

pedrobaeza:
@guewen, there's a return of an empty dictionary 5 lines below, so that's why I say the return value is not consistent. Anyway, if there's no problem with it, go ahead, but don't tell me that "it's consistent"

PR #157, opened 2015-10-21
[FIX] invalid character in header value

lmignon:
`base64.encodestring` adds a trailing new line character to the returned string. This character is not allowed in in the httprequest  header. Use base64.b64encode to avoid trailing new line character in the encoded string


guewen:
:+1: 
Thanks!

PR #164, opened 2015-11-04
[ADD] magentoerpconnect_pricelist_currency

hugosantosred:
- This module is intended to import Magento Orders in the same currency that they were created in magento by using price list mappings.


magentoerpconnect_pricelist_currency/model/sale.py
```diff
@@ -0,0 +1,58 @@
...
+    SaleOrderImporter
+
+
+class MagentoSaleOrder(models.Model):
+    _inherit = 'magento.sale.order'
+
+    magento_currency_id = fields.Many2one('res.currency', 'Magento Currency')
+
+
+@magento(replacing=SaleOrderImportMapper)
```
guewen: When you use `@magento(replacing=SaleOrderImportMapper)`, it introduces a compatibility issue.
If another addon does the same thing, only one of them will win and effectively replace `SaleOrderImportMapper`.

As there is extensibility between backend versions (inheritance through parent backends), it is hard to have modularity for one backend version.

Usually when I encounter a similar issue, my process is the following:
1. Can this feature be integrated in the core module? If yes, it simplifies.
2. If not, I introduce a sort of hook in the core, example: https://github.com/OCA/connector-magento/blob/8.0/magentoerpconnect/sale.py#L430-L435, https://github.com/OCA/connector-magento/blob/8.0/magentoerpconnect/sale.py#L733-L735, in your case it could be SalePricelistImportMapper and SalePricelistImporter. Then I replace only the new ConnectorUnit, reducing the risks of conflicts (still possible but anyway if 2 addons want to change the behavior of the import of pricelists, they are maybe not compatible by design).

magentoerpconnect_pricelist_currency/view/product_pricelist_view.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="product_pricelist_mapping_tree_view" model="ir.ui.view">
+            <field name="name">product.pricelist.mapping.tree</field>
+            <field name="model">product.pricelist.mapping</field>
+            <field name="type">tree</field>
```
guewen: `<field name="type">tree</field>` is deprecated

magentoerpconnect_pricelist_currency/view/product_pricelist_view.xml
```diff
@@ -0,0 +1,66 @@
...
+                <tree string="Pricelist Mappings">
+                    <field name="name"/>
+                    <field name="pricelist_ids" widget="one2many_list"/>
+                </tree>
+            </field>
+        </record>
+        <record id="product_pricelist_mapping_form_view" model="ir.ui.view">
+            <field name="name">product.pricelist.mapping.form</field>
+            <field name="model">product.pricelist.mapping</field>
+            <field name="type">form</field>
```
guewen: `<field name="type">form</field>` is deprecated

magentoerpconnect_pricelist_currency/model/product_pricelist.py
```diff
@@ -0,0 +1,48 @@
...
+    name = fields.Char('Mapping Name', required=True)
+    pricelist_ids = fields.One2many('product.pricelist', 'mapping_id',
+                                    'Mapped Pricelists')
+
+
+class ProductPricelist(models.Model):
+    _inherit = 'product.pricelist'
+
+    mapping_id = fields.Many2one('product.pricelist.mapping',
+                                 'Pricelist Mapping', select=True)
```
guewen: s/select/index/

magentoerpconnect_pricelist_currency/model/sale.py
```diff
@@ -0,0 +1,58 @@
...
+@magento(replacing=SaleOrderImporter)
+class SaleOrderPricelistImporter(SaleOrderImporter):
+    _model_name = ['magento.sale.order']
+
+    def _after_import(self, binding):
+        super(SaleOrderPricelistImporter, self)._after_import(binding)
+        if binding.pricelist_id.currency_id != binding.magento_currency_id:
+            pricelist = binding.pricelist_id.get_pricelist_for_currency(
+                binding.magento_currency_id.id)
+            binding.write({'pricelist_id': pricelist.id})
```
guewen, line 59: Is there a reason for that to be in `_after_import` and not in the Mapper?

magentoerpconnect_pricelist_currency/model/sale.py
guewen: Perhaps you could `replace` https://github.com/OCA/connector-magento/blob/8.0/magentoerpconnect/sale.py#L942-L948, but then your module would be incompatible with `magentoerpconnect_pricing` ( see https://github.com/OCA/connector-magento/blob/8.0/magentoerpconnect_pricing/sale.py#L27-L41)

guewen:
Hi @hugosantosred 

Your addon is interesting and very complete (tests! :smile:)
I just have a concern with the usage of `replacing` which I commented in the diff.

Thanks for your PR

guewen:
(Travis build is red due to an issue with the docs build I'm currently investigating https://github.com/OCA/maintainer-quality-tools/pull/288#issuecomment-153738843)

hugosantosred:
@guewen Thanks for the review! I will made the changes that you comment as soon as I can.

hugosantosred:
@guewen Changes Done! I've introduced hooks in magentoerpconnect core for currency mapping and another in _after_import for pricelist selection. 

I have to do pricelist selection, based on currency, in _after_import method because I need to know the final selected currency in Odoo Sale (Can be different than base sale price list if customer has one selected).

magentoerpconnect_pricelist_currency/model/product_pricelist.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.multi
+    def get_pricelist_for_currency(self, currency_id):
+        self.ensure_one()
+        if not self.mapping_id:
+            return self
+        currency_pricelist = self.search([
+            ('currency_id', '=', currency_id),
+            ('mapping_id', '=', self.mapping_id.id)
+        ])
```
guewen: just spot one thing here: you could add `limit=1`, it also simplifies the next line to:

``` python
return currency_pricelist or self
```

guewen:
Thanks! I just added one last comment, apart that it sounds good now

guewen:
:+1:

PR #168, opened 2015-11-12
[Fix] Use base prices by default when importing magento sales

hugosantosred:
This PR is intended to fix a problem caused by the prices that are imported in Odoo sale order when the magento sales order is made in another currency than base store currency.

Without this commit this prices are imported mixing base_\* prices and magento order prices, so if the sales order imported from magento is made in another currency than the magento store default currency the order prices and the base_\* prices doesn't match. For example the base_shipping_amount in magento is 10€ but the shipping_amount is 11,04$

I think that in magentoerpconnect base module the base_\* prices have to be used in order to avoid not matching amounts from magento sales order and from odoo sale order.


guewen:
Thanks! :+1: 

Build is failing due to the docs, but lint and tests are green :green_apple:

liebana:
Tests in our side are OK, this looks good to me. :+1:

guewen:
need a rebase

guewen:
@hugosantosred could you have a look on https://github.com/OCA/connector-magento/pull/149 which seems closely related?

PR #174, opened 2015-11-17
[8.0] Migration of magentorepconnect_order_comment

hugosantosred:
Migrated module magentoerpconnect_order_comment to version 8.0 with new API

### DIFF
--- a/magentoerpconnect_order_comment/README.rst
+++ b/magentoerpconnect_order_comment/README.rst
@@ -1,0 +1,62 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License
+
+Sale order comments synchronisation
+===================================
+
+Extension for **Magento Connector**
+
+Features
+--------
+
+* Import sale comments in the same time than 'sale order'
+* Move messages from canceled order to replacing sale order (edited sale
+  order in magento)
+* Export messages from OpenERP sale order to Magento comment
+* Export of moved messages to Magento
+
+Settings / options
+------------------
+
+* Ability to require Magento to send email to customer for each comment
+  received. Unactive by default (Connectors/Magento/Store menu).
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/107/8.0
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/connector-magento/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed `feedback
+<https://github.com/OCA/
+connector-magento/issues/new?body=module:%20
+magentoerpconnect_order_comment%0Aversion:%20
+8.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+Credits
+=======
+
+Contributors
+------------
+
+* David BEAL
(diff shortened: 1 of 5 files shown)


magentoerpconnect_order_comment/magento_model.py
```diff
@@ -18,20 +18,14 @@
...
-                   "based on 'send a message' case (not 'log a note')")),
-    }
-
-    _defaults = {
-        'send_sale_comment_mail': False,
-    }
+    send_sale_comment_mail = fields.Boolean(
+        'Send email notification on sale comment',
+        help=_("Require Magento to send email on 'sale order comment' "
+               "based on 'send a message' case (not 'log a note')"),
```
guewen: Help messages are automatically translated, you should not use `_()` on them.
(I know it was wrong before but it is the right opportunity to correct it)

guewen:
Only a small remark regarding a correction with `_()`, apart that, that's fine!

Thanks
:+1:

hugosantosred:
I didn't know that about help on fields and translations. Thanks! @guewen

magentoerpconnect_order_comment/sale.py
```diff
@@ -96,66 +91,47 @@ class magento_sale_comment(orm.Model):
...
+    backend_id = fields.Many2one(
+        'magento.backend', related='magento_sale_order_id.backend_id',
+        string='Magento Backend',
+        store=True,
+        readonly=True)
+    storeid = fields.Char(
+        'Store id',
+        help=MAGENTO_HELP)
+
+    @api.model
```
StefanRijnhart, line 124: ~~On the create method, you need to apply the @api.returns('self', lambda value: value.id) decorator for backwards compatibility with old API overrides of this method (you know, just in case...)~~

magentoerpconnect_order_comment/README.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Sale order comments synchronisation
+===================================
+
+Extension for **Magento Connector**
+
+Features
+--------
+
+* Import sale comments in the same time than 'sale order'
```
StefanRijnhart, line 12: _at the same time as_

magentoerpconnect_order_comment/README.rst
```diff
@@ -0,0 +1,62 @@
...
+* Move messages from canceled order to replacing sale order (edited sale
+  order in magento)
+* Export messages from OpenERP sale order to Magento comment
+* Export of moved messages to Magento
+
+Settings / options
+------------------
+
+* Ability to require Magento to send email to customer for each comment
+  received. Unactive by default (Connectors/Magento/Store menu).
```
StefanRijnhart, line 22: _Inactive_

StefanRijnhart:
Can you catch ImportError for the import of bs4 and then at runtime (in a method where you use the import) check if bs4 is defined, and raise then?

```
try:
    from bs4 import BeautifulSoup
except ImportError:
    BeautifulSoup = False

...

    if not BeautifulSoup:
        raise ....
    return {'comment': BeautifulSoup(comment).get_text()}
```

magentoerpconnect_order_comment/sale.py
guewen, line 124: I was sure it was inherited. This comment would tend to confirm that, but it could be wrong https://github.com/odoo/odoo/blob/9.0/openerp/api.py#L77
Had you ever had issues with that?

StefanRijnhart:
You seem to miss https://github.com/OCA/connector-magento/commit/0ead0327ecc2, and I think you would need to take the i18n directory as well.

magentoerpconnect_order_comment/sale.py
StefanRijnhart, line 124: You are right: I can't seem to reproduce any such issues with a quick test. Thanks for the pointer!

PR #176, opened 2015-11-26
IMP. Account analytic account and Fiscal Position on Sale Order.

lmignon:
It's now possible to specify the analytic account and the fiscal position to use when importing a sale order from magento. Theses options can be specified at any levels of the hierarchy of the backend models according to the needs. In  this way it become possible for exemple to specify a fiscal position to use for a specific store and take advantage of the taxes mapping defined for this specific position to map the right taxe on the sale order line

### DIFF
--- a/magentoerpconnect/magento_model.py
+++ b/magentoerpconnect/magento_model.py
@@ -166,6 +166,21 @@ class MagentoBackend(models.Model):
         string='Magento Products',
         readonly=True,
     )
+    account_analytic_id = fields.Many2one(
+        comodel_name='account.analytic.account',
+        string='Analytic account',
+        help='If specified, this analytic account will be used to fill the '
+        'field  on the sale order created by the connector. The value can '
+        'also be specified on website or the store or the store view.'
+    )
+    fiscal_position_id = fields.Many2one(
+        comodel_name='account.fiscal.position',
+        string='Fiscal position',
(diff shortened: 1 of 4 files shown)


liebana:
Why account analytc + fiscal position at the same time?

lmignon:
@liebana The fiscal position define the way to fill the taxes on the lines. One of our clients has defined different stores on his Magento to sell products in different countries. The customer is redirected to the right store according to its IP.  When we import a sale order into Odoo we must use the right taxes according to the country where the sale has been made. To do that when the sale order is imported into Odoo we assign the fiscal position defined on the Magento store to the new sale order. The fiscal position defines the mapping between the original tax on the product and the one to use for this specific position according to the fiscal rules applicable for the country where the sell is made.

The analytic account is a way to group our sales into a specific analytic account.  In my case it's used to give a way to our client to analyse the financial result of each store individually.

liebana:
Thanks for the explanation, we'll for sure test it. But IMHO this should be separate modules, I may need fiscal position mappings but not the one of analytic account.

lmignon:
@liebana Both are optional. You are free to specify if you want to use one or the other or both -> no need to write separate modules.

guewen:
:+1:

laetitia-gangloff:
:+1:

lmignon:
Thank your @guewen !

PR #185, opened 2015-12-15
[FIX] Keep advisory lock only if we need to update data into Odoo

lmignon:
This change is required to avoid to keep lock when data are already up to date.


guewen:
LGTM :+1:

damdam-s:
:+1:

PR #194, opened 2016-01-15
[fix] compute function of many2one now returns record not int

gfcapalbo:
Changed the computed field to return the record not the .id 
this bug mades the module uninstallable, after change works fine.


guewen:
Thanks! :+1:

StefanRijnhart:
:+1:

PR #202, opened 2016-03-18
[FIX] Avoid overwrite name field over a company customer

kikopeiro:
When you are importing an update of an company customer the "company" field is not saved properly in field "name".

### DIFF
--- a/magentoerpconnect/partner.py
+++ b/magentoerpconnect/partner.py
@@ -579,6 +579,16 @@ class AddressImporter(MagentoImporter):
             return self.address_infos.magento_record
         else:
             return super(AddressImporter, self)._get_magento_data()
+    
+    def _clean_fields_company(self, data):
+        """ Avoid overwrite name field when we import a company. """
+        if data.get('is_default_billing', False) and data.get('company', False):
+            data['name'] = data['company']
+        return data
+
+    def _update(self, binding, data):
+        data = self._clean_fields_company(data)
+        return super(AddressImporter, self)._update(binding, data)
 
     def _define_partner_relationship(self, data):
         """ Link address with partner or parent company. """


guewen:
Hi,

Thanks for your PR!

Could you give an example? With the current behaviour and your expected result.

kikopeiro:
We have already created a company with magento_connector on Odoo.

This customer is configurated has follows in Magento...
Customer information: there is not much to say
Customer addess: is_billing_default --> True, company_field --> filled

The customer in ERP has been created has follows:
Customer information --> name --> the information in Magento "company" field in the address object

If latter we update the customer in Magento,the information when we import to Odoo is not properly setted

Customer information: name --> the information in Magento "name" field in the address object

Thanks

guewen:
Ok, I think I get it.
It should be corrected in the mapper, did you identify why the mapper set it the first time and not the second one?

kikopeiro:
Mapper seems is working fine, **_get_address_infos** function in PartnerAddressBook. Here the code set the company field in customer name on Odoo

```
                company_mapper = self.unit_for(CompanyImportMapper,
                                               model='magento.res.partner')
                map_record = company_mapper.map_record(magento_record)
                parent = partner_binding.openerp_id.parent_id
                values = map_record.values(parent_partner=parent)
                **partner_binding.write(values)**
```

but after this "write" the partner information is being overwritten while executes function "run" for the address import.

```
    **addresses = self._get_address_infos(magento_partner_id,
                                        partner_binding_id)**
    for address_id, infos in addresses:
        importer = self.unit_for(MagentoImporter)
        **importer.run(address_id, address_infos=infos)**
```

guewen:
Still, you should probably not need to modify the importer, the address mapper could return the correct name at first place.

PR #210, opened 2016-05-19
[WIP/MIG][9.0] Migrate magentoerpconnect

SodexisTeam:
Needs label : **Work in progress**

Changes done:
- Obtained the histories from 8.0 using https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0
- Checked pep8 conventions using oca-autopep8
- Version and dependency change
- Changed to simple headers
- Modified README.rst
- Add import for virtual and downloadable products as service and digital respectively from PR https://github.com/OCA/connector-magento/pull/214


magentoerpconnect/product_category.py
```diff
@@ -265,7 +250,7 @@ def parent_id(self, record):
         mag_cat_id = binder.to_openerp(record['parent_id'])
 
         if category_id is None:
-            raise MappingError("The product category with "
-                               "magento id %s is not imported." %
+            raise MappingError(_("The product category with "
+                                 "magento id %s is not imported.") %
```
guewen: Those are technical errors, not meant to be translated (same comment at the other places).

guewen:
Could be nice to use https://github.com/kevin1024/vcrpy for the test data.

guewen:
Exceptions of type `UserError` are shown to the end user and should be translated. I'm really not fond of translatable error messages for the low-level errors such as the network or mapping errors.

PhilippeBach:
a little suggest for the V9   for categories could we import the categories in the same fields than odoo Ecommerce? If one time you want to develop odoo frontend  you dont need to make them one time more 

my post could be a a wrong place but there is no real board for discutions and suggest for OCA 

Regards 

Philippe

magentoerpconnect/product_category.py
atchuthan: We have similar errors in community and odoo modules that contain translation _(...) for technical errors.

guewen: What I mean by technical errors is that they are not supposed to be shown to the users with a `UserError` popup.

I distinguish "error messages" with are meant entirely for the users and should be localized and "exceptions messages" that must not be translated. These exceptions messages are for developers and not end-users. This is important to be able to search for errors and in some case to analyze the content of the exception in `try: except:`.

atchuthan: thanks for your info... will remove translation for technical errors

atchuthan: For info, in the lint check for Travis, it tests for error raised with translations.

magentoerpconnect/sale.py:359: [C8107(translation-required), SaleImportRule._rule_authorized] String parameter of raise "OrderImportRuleRetry" requires translation. Use _('The order has not been authorized.\nThe import will be retried later.')

Gonna ignore this case for technical errors.

StefanRijnhart:
What is the status of this? The commit history and the diff looks mangled, with commits by transifex etc.

atchuthan:
@StefanRijnhart rebase done

StefanRijnhart:
Thanks for the rebase. Can you check the output of the travis jobs?

berpweb:
Any updates with this PR? I restarted the runbot which looks fine now. Travis 👎

berpweb:
Someone working on this or just an abandoned PR?
Ping: @SodexisTeam 

@guewen #255 Is this PR supersede the current one here for V9?

guewen:
@berpweb I'm working on the v10.0 on #255.

PR #212, opened 2016-05-20
8.0 Add support for multi company

lmignon:
The goal of this PR is to add the ability to support multi companies with the connector Magento. This PR follows the same idea as the one expressed in the blueprint of @guewen  https://blueprints.launchpad.net/openerp-connector-magento/+spec/import-multi-company multi company imp
At this stage multi company is supported for
- [X] sale.order: the sale.order are created by company
- [X] res.partner: the re.partner are created without  company if the backend is configured in multi company mode.

In our case, we don't need to import the products from magento, therefore, nothing has  been modified to support the mutli company on products.
- [ ] product.product

By default nothing is provided by magento to support different warehouses by website/store/storeview. In our case we use a specific addon for magento (from smile) and we have developped a specialized odoo addon to export the product qty by warehouse.
- [ ] stock???

### DIFF
--- a/magentoerpconnect/magento_model.py
+++ b/magentoerpconnect/magento_model.py
@@ -181,6 +181,12 @@ class MagentoBackend(models.Model):
         'The value can also be specified on website or the store or the '
         'store view.'
     )
+    is_multi_company = fields.Boolean(
+        string='Is Backend Multi-Company',
+        help="If this flag is set, it is possible to choose warehouse at each "
+             "level. "
+             "When import partner, ignore company_id if this flag is set.",
+    )
 
     _sql_constraints = [
         ('sale_prefix_uniq', 'unique(sale_prefix)',
--- a/magentoerpconnect/magento_model.py
+++ b/magentoerpconnect/magento_model.py
@@ -370,6 +370,15 @@ class MagentoConfigSpecializer(models.AbstractModel):
         'The value can also be specified on website or the store or the '
         'store view.'
     )
+    specific_warehouse_id = fields.Many2one(
+        comodel_name='stock.warehouse',
+        string='Specific warehouse',
+        help='If specified, this warehouse will be used to load fill the '
+        'field warehouse (and company) on the sale order created by the '
+        'connector.'
+        'The value can also be specified on website or the store or the '
+        'store view.'
+    )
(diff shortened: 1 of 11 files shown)


StefanRijnhart:
Hi @lmignon, @laetitia-gangloff, is this perhaps ready for review?

lmignon:
@StefanRijnhart This PR provides a partial solution to support multi companies with magento. This code is used in production. IMO it's ready for review...

StefanRijnhart:
We have been testing this (we can live with the limitations). :+1: and thanks!

magentoerpconnect/magento_model.py
```diff
@@ -364,6 +370,15 @@ class MagentoConfigSpecializer(models.AbstractModel):
         'The value can also be specified on website or the store or the '
         'store view.'
     )
+    specific_warehouse_id = fields.Many2one(
+        comodel_name='stock.warehouse',
+        string='Specific warehouse',
+        help='If specified, this warehouse will be used to load fill the '
+        'field warehouse (and company) on the sale order created by the '
+        'connector.'
```
StefanRijnhart, line 378: This is not working in my setup. We get another warehouse from the same company. In fact, I cannot seem to find any place (in this patch or elsewhere) where the Magento warehouse setting is used to assign a warehouse to the sale orders imported from Magento. The Magento documentation does not even claim that the warehouse setting is used as such, but says the warehouse is used to gather the available product quantities instead (http://odoo-magento-connector.com/howto/configure_warehouse.html). Can you confirm that this setting is actually ignored for sale orders created from the Magento import?

lmignon:
@StefanRijnhart Indeed, we only use this information to compute the stock for a specific warehouse into magento (custom module in magento)

StefanRijnhart:
@lmignon How about https://github.com/acsone/connector-magento/pull/1?

susport (commented):
Please to corrrect de travis error.

StefanRijnhart:
@susport those errors are unrelated to this change and can be fixed separately.

StefanRijnhart (approved):
Been running this in production for months now.

Cedric-Pigeon:
@guewen Can you consider this PR in your V10 migration ?

dreispt (approved):

PR #20, opened 2014-07-23
Module to map tax product

albertofl:
Migraton for bazaar

PR #213, opened 2016-05-24
[IMP] Add module to map product taxes.

jesusVMayor:
I used @albertofl module: https://github.com/OCA/connector-magento/pull/20


StefanRijnhart:
Thanks. Can you follow up the comments from the original PR? Notably, apply a decorator with the replace argument and create a more useful description. Please also provide a README.rst instead of the description in __openerp__.py (see https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst).

PR #214, opened 2016-05-24
[8.0]Add import for virtual and downloadable products as services.

jesusVMayor:


atchuthan:
:+1:

magentoerpconnect/product.py
```diff
@@ -71,11 +71,11 @@ def product_type_get(self):
         return [
             ('simple', 'Simple Product'),
             ('configurable', 'Configurable Product'),
+            ('virtual', 'Virtual Product'),
+            ('downloadable', 'Downloadable Product'),
```
atchuthan: instead of downloadable, you could use "digital"  [Suggestion]

guewen:
:+1:

l-arnold:
+1 
As a short term workaround, I have found that I can build first as Virtual in Magento , then go in and change to Simple in MySQL, then sync to Odoo, then change back to Virtual in Magento. 

Setting a Service in Odoo is a good idea!

StefanRijnhart:
:+1: Merging this as it looks like the CI breakage is unrelated.

PR #215, opened 2016-06-29
[IMP] Set addresses of guests to False as well

StefanRijnhart:
It seems arbitrary to have guests' addresses as active in the system, as opposed to logged-in customers. The duplication (there are always two additional addresses apart from the partner itself) is still pretty annoying.

### DIFF
--- a/magentoerpconnect/sale.py
+++ b/magentoerpconnect/sale.py
@@ -872,13 +872,10 @@ class SaleOrderImporter(MagentoImporter):
         # the partner form or the searches. Too many adresses would
         # be displayed.
         # They are never synchronized.
-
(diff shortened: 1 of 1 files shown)


StefanRijnhart:
BTW I don't understand the address book comment. cc original author @guewen

guewen:
Totally agree with this change.
The original comment was well in v7 when we had partner and addresses: we didn't wanted a partner without addresses. There is no reason to keep this behaviour in post-v7.
:+1:

atchuthan:
:+1:

pedrobaeza:
:+1:

StefanRijnhart:
Thanks for the clarification and the quick response everyone!

PR #218, opened 2016-07-05
typo on field used in ecommerce to manage giftcode

damdam-s:
to be coherent with https://github.com/OCA/connector-ecommerce/blob/8.0/connector_ecommerce/sale.py#L322-L323


guewen:
:+1:

StefanRijnhart:
:+1:, thanks!

damdam-s:
up ! 
I added a control to avoid creating SO line with `amount  = 0` if no gift is in the magento order.

atchuthan:
:+1:

PR #219, opened 2016-07-05
[IMP] Bind imported products to existing ones with the same code

StefanRijnhart:


StefanRijnhart:
If there is a reason that this not implemented yet, please enlighten me!

guewen:
There were discussions and when people asked if it could be supported, I said more than once I would accept such a pull request but nobody did it, until you :)

Thanks!
:+1:

PhilippeBach:
Congratulation i dreamed for

l-arnold:
Excellent direction.  Will do a pull.  There are many times where it would be efficient to first add a product in Odoo then get it linked in Magento later.  Would also be nice to combine products in Odoo.  If a product is listed in multiple Magento websites it makes multiple Odoo products in my understanding and from Business Managment, it would generally be one Product with multiple locations (or not even)

StefanRijnhart:
@guewen great, thanks for the quick response! And thank you for the great connector framework that makes all of this so easy and elegant to achieve.

atchuthan:
:+1: great!

Could we also add an option in the Connector/Magento/backend "fail_on_code_mismatch"(boolean) such that the job fails if the product code does not match?

liebana:
Internal code is not unique by default in Odoo, but it's in Magento. How would you deal if the code is repeated in more than one product? This could be a common case in a multi-company enviroment where products are not shared among them.

StefanRijnhart:
@liebana do you mean that the change in behaviour in this PR should be optional?

liebana:
To avoid conflicts, I think this behaviour should be linked to a new functionality in order to not allow repeated internal references in Odoo, as Magento does with their SKUs.

StefanRijnhart:
@liebana to be fair, that is actually not trivial if you take the visibility of products across companies into account. For the more trivial cases, there is always https://github.com/OCA/product-attribute/tree/8.0/product_sequence but I don't think the uniqueness constraint should be required.

StefanRijnhart:
@guewen do you think this can be merged now? I count two oldstyle :+1:

PR #220, opened 2016-07-20
[FIX][9.0] magentoerpconnect: Assertion for zero record

lasley:
This fixes the zero record assertion, please see below example which clarifies edge case. Intent of this method was to fail on a False, but that is not currently the case

```
>>> external_id = False
>>> binding_id = True
>>> assert (external_id or external_id == 0) and binding_id
# No assertion raised
```

My fix seems to hit the intent:

```
>>> assert (external_id or not isinstance(external_id, int)) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
```

I'll create a PR for this for v8 as well too, just figured we would rock out one at a time in case there is a better suggestion or I am being dumb in the intent inference.

### DIFF
--- a/magentoerpconnect/unit/binder.py
+++ b/magentoerpconnect/unit/binder.py
@@ -119,7 +119,9 @@ class MagentoModelBinder(MagentoBinder):
         """
         # the external ID can be 0 on Magento! Prevent False values
         # like False, None, or "", but not 0.
-        assert (external_id or external_id == 0) and binding_id, (
+        assert (
+            external_id or not isinstance(external_id, int)
+        ) and binding_id, (
             "external_id or binding_id missing, "
             "got: %s, %s" % (external_id, binding_id)
         )
--- a/magentoerpconnect/unit/binder.py
+++ b/magentoerpconnect/unit/binder.py
@@ -119,9 +119,7 @@ class MagentoModelBinder(MagentoBinder):
         """
         # the external ID can be 0 on Magento! Prevent False values
         # like False, None, or "", but not 0.
-        assert (
-            external_id or not isinstance(external_id, int)
-        ) and binding_id, (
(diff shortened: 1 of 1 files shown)


guewen:
Hi,

Thanks for your proposal.

`0` would no longer be a valid value:

``` python
>>> binding_id = True
>>> external_id = 0
>>> assert (external_id or not isinstance(external_id, int)) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
```

Perhaps your intent was to write:

``` python
assert (external_id or isinstance(external_id, int)) and binding_id
```

But still, this is no better:

``` python
>>> binding_id = True
>>> external_id = False
>>> assert (external_id or isinstance(external_id, int)) and binding_id
```

This is because `False` and `True` are nothing more than `0` and `1`:

``` python
>>> False == 0
True
>>> True == 1
True
```

I can think of doing:

``` python
assert (external_id or external_id == 0 and external_id is not False) and binding_id
```

Which is not really a beauty but should cover all the cases.

lasley:
Damn, you're right. I originally had the last one of your's, but I agree that it is sloppy. What about:

``` python
assert (external_id or external_id is 0) and binding_id
```

Values:

``` python
>>> external_id = 'test'
>>> assert (external_id or external_id is 0) and binding_id
>>> external_id = 0
>>> assert (external_id or external_id is 0) and binding_id
>>> external_id = False
>>> assert (external_id or external_id is 0) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
>>> external_id = None
>>> assert (external_id or external_id is 0) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
```

guewen:
Nice one! :+1:

StefanRijnhart:
:+1:

PR #221, opened 2016-07-21
[FIX][8.0] magentoerpconnect: Assertion for zero record

lasley:
Seems like we came up with the good fix for this in #220, so this is the identical PR for 8.0 before I forget to submit it!


StefanRijnhart:
Merging as per #220

PR #222, opened 2016-07-21
[FIX][7.0] magentoerpconnect: Assertion for zero record

lasley:
Last but not least, this is the 7.0 PR for the zero record assertion fix described in #220 

For future reference, was there a more efficient way for me to do this whole bug fix across three versions thing? Seems like this is something that should be semi-automated, or am I just dreaming?


StefanRijnhart:
Semi-automatic forward porting for larger changes than this one is non-trivial. It would put the burden of resolving all conflicts with a maintainer that was not the author. Conflicts might have to be split up per change, which is time consuming. The changes would require review, just like this one or the process would be error prone. Because of the review it would not be very automatic. In short, thank you for taking the time to prepare the backports!

StefanRijnhart:
Merged as per the review on #220.

PR #230, opened 2016-08-17
[RFR] Make pushing prices to Magento optional

StefanRijnhart:
so that pricelists can also be used to just import orders in different
currencies per website.


guewen:
Great! Looking at the resulting code, an option is definitively better :)

:+1:

PR #237, opened 2016-09-29
Better assignment of main category

gfcapalbo:
When fetching categories from magento the connector returns a list of category ids.
To decide which one will be updated as categ_id  (main category)  the connector pops the last element of the categories and elects it as the main category of the product.
This randomness may cause unpredictable behaviour on update when using multiple categories in magento.
We had a client who wanted some products to be category:  product_price/ 6 and product_price 5.
when updating products the connector would substitute some of the categories of some products and leave other intact.  (depending on wich category happened to be the last in the magento category list.)

We are adding a little bit more logic to this decision. If the current main category of the odoo product exists already in the list of returned categories from magento, do not update the main category at all.

### DIFF
--- a/magentoerpconnect/product.py
+++ b/magentoerpconnect/product.py
@@ -468,13 +468,10 @@ class ProductImportMapper(ImportMapper):
     def categories(self, record):
         mag_categories = record['categories']
         binder = self.binder_for('magento.product.category')
-        curr_product = self.binder_for('magento.product.product').to_openerp(
-            record['product_id'], unwrap=True
-        )
-        curr_product_rec = self.env['product.product'].browse(curr_product)
-        category_ids = []
+        product_id=[]
+	curr_product = self.binder_for('magento.product.product').to_openerp(record['product_id'], browse=True)
+	category_ids = []
         main_categ_id = None
-
         for mag_category_id in mag_categories:
             cat_id = binder.to_openerp(mag_category_id, unwrap=True)
             if cat_id is None:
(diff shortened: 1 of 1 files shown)


magentoerpconnect/product.py
```diff
@@ -491,8 +491,14 @@ def categories(self, record):
...
         result = {'categ_ids': [(6, 0, category_ids)]}
-        if main_categ_id:  # OpenERP assign 'All Products' if not specified
-            result['categ_id'] = main_categ_id
+        # OpenERP assign 'All Products' if not specified
+        # skip main cat assignment if the current main category is already in
+        # categ_ids
+        if main_categ_id:
+            if self.backend_record.categ_id.id not in category_ids):
+                result['categ_id'] = main_categ_id
+            else:
```
hbrunn: why would you need this? This branch causes the value to be overwritten with itself

magentoerpconnect/product.py
```diff
@@ -491,8 +491,14 @@ def categories(self, record):
                 main_categ_id = default_categ.id
 
         result = {'categ_ids': [(6, 0, category_ids)]}
-        if main_categ_id:  # OpenERP assign 'All Products' if not specified
-            result['categ_id'] = main_categ_id
+        # OpenERP assign 'All Products' if not specified
+        # skip main cat assignment if the current main category is already in
+        # categ_ids
+        if main_categ_id:
+            if self.backend_record.categ_id.id not in category_ids):
```
hbrunn: python syntax error

hbrunn (changes requested):

PR #255, opened 2017-06-16
[10.0] migrate connector_magento

guewen:
Depends on:
* https://github.com/OCA/connector/pull/231 (merged)
* https://github.com/OCA/connector-ecommerce/pull/41 (merged)
* https://github.com/OCA/server-tools/pull/858 (merged)

* [x] main API migration for `magentoerpconnect`
* [x] update tests
* [x] update docs


pedrobaeza:
There are conflicts right now.

Shouldn't this be better renamed to connector-magento?

guewen:
> There are conflicts right now.

Yep, I'll fix them, but my focus is elsewhere at this moment, will do it soon.

> Shouldn't this be better renamed to connector-magento?

It would be nice, and the right moment to do it, anyway addons that depend on it need changes.

Cedric-Pigeon:
@guewen Hi what is the status of this PR ? Is it usable ?

guewen:
Ready for reviews!

server_env_connector_magento/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2014-2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{'name': 'Magento Connector - Server Environment Configuration',
+ 'version': '10.1.0.0',
```
Cedric-Pigeon: Wrong version number structure, should have 5 numbers

Cedric-Pigeon (changes requested):

connector_magento/__manifest__.py
```diff
@@ -0,0 +1,42 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{'name': 'Magento Connector',
+ 'version': '10.0.1.0.0',
+ 'category': 'Connector',
+ 'depends': ['account',
+             'product',
+             'delivery',
+             'sale_stock',
+             'connector_ecommerce',
```
Cedric-Pigeon: @guewen Why did you remove the depends on product_multi_category ?

Cedric-Pigeon (changes requested):

connector_magento/__manifest__.py
guewen: well spotted, I thought it was unused, turns out I was wrong :)

Cedric-Pigeon:
@guewen To use VCR in your tests suite, did you use a tool to convert python dict in xml in yaml file ?

guewen:
@Cedric-Pigeon No, I rerun the tests on a Magento instance

connector_magento/models/sale_order/common.py
```diff
@@ -0,0 +1,295 @@
...
+    _name = 'magento.sale.order.line'
+    _inherit = 'magento.binding'
+    _description = 'Magento Sale Order Line'
+    _inherits = {'sale.order.line': 'odoo_id'}
+
+    magento_order_id = fields.Many2one(comodel_name='magento.sale.order',
+                                       string='Magento Sale Order',
+                                       required=True,
+                                       ondelete='cascade',
+                                       select=True)
```
Cedric-Pigeon: You should use `index=` in place of `select=`

geluboros:
Hi @guewen,
Great work done here! I tested the module and everything looks great.

Just one small thing:
After importing 2 orders from magento, when i try to process their picking, i get a cursor error on the second picking. If i refresh instance/module the process works again for the first picking but fails on the second one (and so on).

**Error:** OperationalError: Unable to use a closed cursor. 

I think the error is related to this issue odoo/odoo#6597

guewen:
@gboros-rgbconsulting Thanks for your test and report! I fixed this issue on https://github.com/OCA/connector/pull/231/commits/776ae6d641026bdaf418d60d661d300890cbf717 in the PR https://github.com/OCA/connector/pull/231/

connector_magento/components/backend_adapter.py
```diff
@@ -101,27 +47,84 @@ def location(self):
         return location
```
Cedric-Pigeon, line 49: It seems to be a bug. `use_auth_basic` is always false even if checked on the backend.
I am trying to determine why but it is not clear for me.

Cedric-Pigeon: @guewen I found it out and just made a PR on your branch. Could you please have a look ?
https://github.com/guewen/connector-magento/pull/7

guewen:
@Cedric-Pigeon merged, thanks!

jaredkipe:
Thank you very much for the work on this @guewen !

I think it would be nice to have "Default Product Category" assign that as the main category on product import regardless of whether or not there is a category on the product in Magento.  Otherwise a lot of duplicate setup/config would need to happen to setup effectively all of the categories to have correct inventory valuation etc. OR the end user needs to change the category and potentially put the same "other" category into the "other categories" m2m.

Fiscal positions are good, but what do you think about analytic accounts?

I'm testing currently on a Magento store with thousands of products and hundreds of orders so I'm still importing products... Is there anything I can do other than lowering the runner count to 1 to prevent "concurrent update" failures?  (this may be a question for the OCA/queue repo, but I know you have a lot of experience with it)

jaredkipe:
Fails to import 'bundle' products, though there appears to be a class to do this (or at least let you customize it).

Would you accept a patch to throw 'giftcard' in the tuple of product types that map to service type?

jaredkipe:
@guewen 
~~I'm having a hard time using the Component _inherit to do something really simple like patch over ProductImportMapper.type() to include 'giftcard'.~~

~~From the docs,  http://connector-rework-doc.surge.sh/guides/migration_guide.html#migration-guide
, it would seem that just like a typical Odoo patch job, I could make a module, depend on `connector_magento`, define a class that extends Component, and `_inherit = 'magento.product.product.import.mapper'` and redefine the method (with the `@mapping`) presumably.~~

~~This does not seem to work.~~

~~I can tell the Component registry see's my code by doing things like making `_name = 'magento.product.product.import.mapper'` and having the Component registry complain that the name is already defined.  (this would seem to imply that my class is getting added to the registry, and in the correct order).~~

~~Any advice on extending the mapping?~~

Nevermind, after debugging the component registry to the point where I was convinced that it was working correctly I realized that mapping doesn't even happen until the product is validated by a normal odoo model.  After extending that to allow my product type, everything else I was working on with the 'mapper' worked properly.

guewen:
> I think it would be nice to have "Default Product Category" assign that as the main category on product import regardless of whether or not there is a category on the product in Magento. Otherwise a lot of duplicate setup/config would need to happen to setup effectively all of the categories to have correct inventory valuation etc. OR the end user needs to change the category and potentially put the same "other" category into the "other categories" m2m.
> Fiscal positions are good, but what do you think about analytic accounts?

This is a one-to-one migration, those changes should be proposed in a subsequent pull request.

guewen:
> I'm testing currently on a Magento store with thousands of products and hundreds of orders so I'm still importing products... Is there anything I can do other than lowering the runner count to 1 to prevent "concurrent update" failures? (this may be a question for the OCA/queue repo, but I know you have a lot of experience with it)

Normally this type of error is automatically retried. Possibly after 5 attempts, the job is set to failed. This error happens more when "sub-records" are imported by an importer (like when a sales orders is imported and products were not already there, it has to import them as well, now if 2 jobs import a sales order with the same products, one will fail with concurrent failure), so I tend to think that importing first product categories, partner categories, then products and partners, then sales orders should reduce the number of concurrency failures.

guewen:
> Nevermind, after debugging the component registry to the point where I was convinced that it was working correctly I realized that mapping doesn't even happen until the product is validated by a normal odoo model. After extending that to allow my product type, everything else I was working on with the 'mapper' worked properly.

Glad to read that :)

guewen:
> Fails to import 'bundle' products, though there appears to be a class to do this (or at least let you customize it).

Yes, it was not implemented in the earlier versions neither, but there was a hook for implementing them, I kept the hook even if it makes less sense with the new component API to ease the port of prospective custom addons using it.

Update: I think it was for https://github.com/OCA/connector-magento/pull/117

> 
> Would you accept a patch to throw 'giftcard' in the tuple of product types that map to service type?

Yes

jaredkipe:
>>Would you accept a patch to throw 'giftcard' in the tuple of product types that map to service type?

>Yes

Great, I'll get on that.

While I'm at it, would you accept a pull that does MagentoEnterprise GiftCard, StoreCredit, and Rewards?

I see that the current implementation of gift certificate comes from connector_ecommerce and there isn't necessarily an analog to 'store credit' or 'rewards'...

jaredkipe:
>Normally this type of error is automatically retried. Possibly after 5 attempts, the job is set to failed. This error happens more when "sub-records" are imported by an importer (like when a sales orders is imported and products were not already there, it has to import them as well, now if 2 jobs import a sales order with the same products, one will fail with concurrent failure), so I tend to think that ?importing first product categories, partner categories, then products and partners, then sales orders should reduce the number of concurrency failures.

I was only importing products at the time (and had already finished importing categories, so it couldn't have been a race condition for that).  I believe, though I can't be sure, that those jobs were failed and wouldn't be re-tried without manually re-queueing them. 

In the near future I'll be doing all of this again to test some customizations, so I'll try to save screenshots and the error traces.  Its basically the same error I get a lot when running a `-u all` on a database that is also being used by users or at least the cron in another container.  Perhaps I need to adjust some transaction settings in Postgres...

jaredkipe:
After testing, the only problems I've come up with are some deadlocks that do not appear to resolve themselves.

The connector/job_queue broke down with something like this being spammed into the log.

```
odoo_1   | 2017-08-05 22:28:45,632 601 ERROR tests odoo.addons.connector_magento.models.magento_backend.common: Could not acquire advisory lock
odoo_1   | Traceback (most recent call last):
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 229, in synchronize_metadata
odoo_1   |     self.env[model_name].import_batch(backend)
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_binding/common.py", line 44, in import_batch
odoo_1   |     return importer.run(filters=filters)
odoo_1   |   File "/opt/odoo/addons/connector_magento/components/importer.py", line 237, in run
odoo_1   |     self._import_record(record_id)
odoo_1   |   File "/opt/odoo/addons/connector_magento/components/importer.py", line 255, in _import_record
odoo_1   |     self.model.import_record(self.backend_record, external_id)
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_binding/common.py", line 53, in import_record
odoo_1   |     return importer.run(external_id, force=force)
odoo_1   |   File "/opt/odoo/addons/connector_magento/components/importer.py", line 203, in run
odoo_1   |     self.advisory_lock_or_retry(lock_name)
odoo_1   |   File "/opt/odoo/addons/connector/components/core.py", line 130, in advisory_lock_or_retry
odoo_1   |     ignore_retry=True)
odoo_1   | RetryableJobError: Could not acquire advisory lock
odoo_1   | 2017-08-05 22:28:45,633 601 ERROR tests odoo.addons.base.ir.ir_cron: Call of self.env[u'magento.backend']._scheduler_import_product_product(*()) failed in Job #15
odoo_1   | Traceback (most recent call last):
odoo_1   |   File "/usr/local/lib/python2.7/site-packages/odoo/addons/base/ir/ir_cron.py", line 118, in _callback
odoo_1   |     getattr(model, method_name)(*args)
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 345, in _scheduler_import_product_product
odoo_1   |     self._magento_backend('import_product_product', domain=domain)
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 325, in _magento_backend
odoo_1   |     getattr(backends, callback)()
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 300, in import_product_product
odoo_1   |     'import_products_from_date')
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 267, in _import_from_date
odoo_1   |     backend.check_magento_structure()
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 181, in check_magento_structure
odoo_1   |     backend.synchronize_metadata()
odoo_1   |   File "/opt/odoo/addons/connector_magento/models/magento_backend/common.py", line 236, in synchronize_metadata
odoo_1   |     str(e).decode('utf-8', 'ignore'))
odoo_1   | UserError: (u"Check your configuration, we can't get the data. Here is the error:\nCould not acquire advisory lock", None)
```


Without looking too much into it, I simply restarted `odoo` and went on my way.  The next time I came access something like this it was trying to save a store view to have an earlier date to test re-importing something.  Attempting to save a store views locked up the web client, so I went into the database to discover why...

Turns out that there were two existing transactions that were deadlocked on this same row.

```
 blocked_pid | blocked_user | blocking_pid | blocking_user |                                                                   blocked_statement                                                                   |                                                         current_statement_in_blocking_process                                                         
-------------+--------------+--------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------
       19285 | odoo         |        19286 | odoo          | UPDATE "magento_storeview" SET "import_orders_from_date"='2017-08-07 18:12:01',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (19) | UPDATE "magento_storeview" SET "import_orders_from_date"='2017-08-07 18:12:01',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (19)
       19286 | odoo         |        19271 | odoo          | UPDATE "magento_storeview" SET "import_orders_from_date"='2017-08-07 18:12:01',"write_uid"=1,"write_date"=(now() at time zone 'UTC') WHERE id IN (19) | select id from wkf_instance where res_id=48 and res_type='magento.storeview' and state='active'
(2 rows)
```

I needed to `SET lock_timeout = '30s';` to remove them and move on.

Since I followed different steps to diagnose and resolve them, I cannot say if both problems are in fact the same problem, but it is troubling to me.

jaredkipe:
Fiscal positions that are detected automatically detect, and get onto the sale order, but do not re-calculate/map the taxes already on product/shipping lines.

guewen:
Thanks for your detailed reports.

guewen:
> Fiscal positions that are detected automatically detect, and get onto the sale order, but do not re-calculate/map the taxes already on product/shipping lines.

Fixed by dae2d36

guewen:
@jaredkipe you traceback seems to indicate that the backend had no website as it went in `check_magento_structure` > `synchronize_metadata`:

```
    @api.multi
    def check_magento_structure(self):
        """ Used in each data import.

        Verify if a website exists for each backend before starting the import.
        """
        for backend in self:
            websites = backend.website_ids
            if not websites:
                backend.synchronize_metadata()
        return True
```

Is it possible?

The "advisory lock" error means that a job for importing the same record was already running at that time. From my reading, it was triggered by many jobs trying to import the "metadata" at the same time (from `check_magento_structure`). Did you used the "Synchronize Metadata" button before importing products?

jaredkipe:
@guewen, I suppose it is possible.  I'm testing this on more than one Magento store and unfortunately I was not physically the one to add this particular website, so I don't KNOW that they followed the steps correctly.  The more recent one with deadlocked transactions was done in the correct order, all of the metadata had synced.

I'm testing the fiscal position thing now.

Would you like me to add a pull request that adds Enterprise rewards, store credit, and gift card payment lines?

guewen:
> I suppose it is possible. I'm testing this on more than one Magento store and unfortunately I was not physically the one to add this particular website, so I don't KNOW that they followed the steps correctly.

Improvement for the future: have a `state` on the backend and allow sync only when the state is ready, after syncing metadata :)



> The more recent one with deadlocked transactions was done in the correct order, all of the metadata had synced.

I don't get what happened then :-( I never had such issue.



> Would you like me to add a pull request that adds Enterprise rewards, store credit, and gift card payment lines?

Yes please!


I would really :heart: to have this PR merged which would ease further contributions :) (starting by https://github.com/OCA/connector/pull/231) I'm really thankful for your feedbacks :1st_place_medal:

jaredkipe:
@guewen Detected fiscal positions work BTW.  I wonder if that is something that could higher up like in the ecommerce connector, as it would likely be boilerplate for anyone making sales orders.

loopool:
Hi, @guewen
The sync from magento not work. There is a problem with job queue.
Error message in the odoo-server.log
```2017-08-17 08:15:34,773 2904 ERROR odoo10 odoo.http: Exception during JSON request handling.
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 640, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 677, in dispatch
    result = self._call_function(**self.params)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 333, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/service/model.py", line 101, in wrapper
    return f(dbname, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 326, in checked_call
    result = self.endpoint(*a, **kw)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 935, in __call__
    return self.method(*args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/http.py", line 506, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo10/addons/web/controllers/main.py", line 889, in call_button
    action = self._call_kw(model, method, args, {})
  File "/home/odoo/odoo10/addons/web/controllers/main.py", line 877, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/api.py", line 681, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/usr/local/lib/python2.7/dist-packages/odoo-10.0-py2.7.egg/odoo/api.py", line 672, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/home/odoo/odoo10/myaddons/queue/queue_job/models/queue_job.py", line 127, in open_related_action
    action = job.related_action()
  File "/home/odoo/odoo10/myaddons/queue/queue_job/job.py", line 544, in related_action
    return action(**self.func.kwargs)
  File "/home/odoo/odoo10/myaddons/connector-magento/connector_magento/models/queue_job/common.py", line 20, in related_action_magento_link
    external_id = self.args[external_id_pos]
IndexError: list index out of range```

**Comment by guewen** [comment]
> I wonder if that is something that could higher up like in the ecommerce connector, as it would likely be boilerplate for anyone making sales orders.

I agree and thought about that too but didn't find a way to do it with the existing "onchange player", because before calling the `create`, we have some lines in a list of `sale.order.line` and some lines in a list of `magento.sale.order.line`.

**Comment by guewen** [comment]
> The sync from magento not work. There is a problem with job queue.

@loopool 
This error does not happen during execution of a job, but when you click on the button "related action" of a job. Could you show me what the result of a `SELECT * FROM queue_job WHERE id = ?` gives for the job on which you have this error?

**Comment by loopool** [comment]
@guewen 
All my sync job queues have status pending. Below is the sql result.

| id | func_string                                                                                                                                                                        | date_done | result | priority | retry | user_id | uuid                                 | record_ids | message_last_post | company_id | method_name  | state   | kwargs                                                                                                                                           | date_started | channel_method_name               | channel      | args                                                                            | job_function_id | max_retries | date_created        | name                                            | model_name         | eta | date_enqueued | exc_info |
|----|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------|--------|----------|-------|---------|--------------------------------------|------------|-------------------|------------|--------------|---------|--------------------------------------------------------------------------------------------------------------------------------------------------|--------------|-----------------------------------|--------------|---------------------------------------------------------------------------------|-----------------|-------------|---------------------|-------------------------------------------------|--------------------|-----|---------------|----------|
| 33 | magento.sale.order().import_batch(magento.backend(1,), filters={u'from_date': None, u'to_date': datetime.datetime(2017, 8, 17, 9, 13, 57, 117632), u'magento_storeview_id': u'1'}) |           |        | 1        | 0     | 1       | f2647d76-1135-4ee5-a046-3ee52e136bbd | []         |                   | 1          | import_batch | pending | {"filters": {"from_date": null, "to_date": {"_type": "datetime_isoformat", "value": "2017-08-17T09:13:57.117632"}, "magento_storeview_id": "1"}} |              | <magento.sale.order>.import_batch | root.magento | [{"uid": 1, "_type": "odoo_recordset", "model": "magento.backend", "ids": [1]}] | 52              | 5           | 2017-08-17 09:13:57 | Prepare the import of Sales Orders from Magento | magento.sale.order |     |               |          |

**Comment by guewen** [comment]
@loopool Thanks, I fixed the related_action error.

Your job not being run is unrelated though, probably the jobrunner is misconfigured, have a look at http://connector-rework-doc.surge.sh/guides/jobrunner.html

**Comment by jaredkipe** [comment]
> I agree and thought about that too but didn't find a way to do it with the existing "onchange player", because before calling the create, we have some lines in a list of sale.order.line and some lines in a list of magento.sale.order.line.

I tried to troubleshoot it from the perspective of the "onchange player" and couldn't find a solution, which is why I commented about it in the first place.

**Comment by jaredkipe** [comment]
@guewen I'm working on the CLA

How would one go about extending/adding fields to something like SaleOrderImportMapper.direct ?

**Comment by guewen** [comment]
> How would one go about extending/adding fields to something like SaleOrderImportMapper.direct ?

https://github.com/guewen/connector-magento/blob/edc4ce8bc58e7cf568eb9515ddc332c4cffa96c7/connector_magento_customize_example/models/partner.py#L38-L41

For the record:

```python

from odoo.addons.component.core import Component


class PartnerImportMapper(Component):
    _inherit = 'magento.partner.import.mapper'

    @property
    def direct(self):
        mappings = super(PartnerImportMapper, self).direct[:]
        return mappings + [('created_in', 'created_in')]

```

l-arnold:
Small note that likely need to update "docs" in Branch GH-Pages to show "downloadable" and "gift card" types as they are now allowed.

Have other new types entered in the allowable stack?   I wiill work on a test setup for 10.0  in the coming days.

Thanks Guewen and others for you hard work on this!

jaredkipe:
@guewen would you accept a pull request with a very slight override to the product importer to bind to an existing product by SKU?

I could do it as a module, like `connector_magento_product_sku` or something.

guewen:
@jaredkipe yes, would be great

kurniawanlucky:
@guewen hallo, i have a question with connector,
I need attributes at Magento "Manufacturer" For Products (Menu Catalog>Attributes>Manage Attributes> attributes code: manufacturer), 
Magento table: manufacturer
so, You have a solution?
Thank's

jaredkipe:
@guewen Is there anything I can do about the couple of open pull requests I have open?  I have received OCA confirmation on my CLA.

jaredkipe:
@guewen Pickings created via drop shipping do not flow back to Magento.

jaredkipe:
@guewen Every time I import an order, a new 'archived' partner is created for both the partner_invoice_id and partner_shipping_id.  The orders do use an existing partner for the partner_id field, but not for the other two.

The newly created 'archived' partners do not have any bindings, but are children of the main partner_id which does have bindings.

sbidoul:
@jaredkipe can you send your signed CLA?

jaredkipe:
@sbidoul What do you mean? I sent it to cla@odoo-community.org and received an acknowledgement on September 4th.

sbidoul:
@jaredkipe apologies, I misread a previous message. Everything is ok.

sbidoul:
@guewen Except the one failing test, do you think this PR is good enough for merging? That would facilitate further contributions.

guewen:
> @guewen Except the one failing test, do you think this PR is good enough for merging? That would facilitate further contributions.

Yes it is and indeed it would make contributions a lot easier, I have to find some time to fix the test.

guewen:
> Every time I import an order, a new 'archived' partner is created for both the partner_invoice_id and partner_shipping_id. The orders do use an existing partner for the partner_id field, but not for the other two.
> 
> The newly created 'archived' partners do not have any bindings, but are children of the main partner_id which does have bindings.

This is normal: on its side, Magento creates a new address for every order (and the API doesn't return us ids for them), so we have to replicate the same behavior. The addresses are archived so you can use them on the sales orders and related document but not in the searches/lists. This is the difference between the "address book" (the not archived addresses) and the "sales addresses".

jaredkipe:
> This is normal: on its side, Magento creates a new address for every order (and the API doesn't return us ids for them), so we have to replicate the same behavior. The addresses are archived so you can use them on the sales orders and related document but not in the searches/lists. This is the difference between the "address book" (the not archived addresses) and the "sales addresses".

Makes sense.

Regarding the failing test case.  Perhaps I am not reading it correctly, but I can't find anywhere where it would appear that you are only looking for the stock on the location given by ref("stock.stock_location_components") (5.0 qty).

It would seem that ref("stock.stock_location_components") is inside WH (WH/Stock/Shelf 1), so I would expect its inventory to contribute to WH's inventory (and that is the warehouse assigned to the backend in MagentoTestCase.setUp().

If you want the test to pass, you can put this line after line 192 (my_location_id = ...) `binding = binding.with_context(location=my_location_id)`

That makes the test pass, but only because of the implementation of the implementation.  I think changing line 208 to `self.assertEqual(binding.magento_qty, 35.0)` is correct.  

All this said, I'd rather get to select a field (probably optional) for a location, falling back to the WH location if not set.  This would allow you to combine or filter stock going out based on views or physical locations. (e.g. if you have more than one warehouse you can send the stock location 'Physical Locations' to send all of them combined)

connector_magento/components/exporter.py
```diff
@@ -73,62 +53,58 @@ def _delay_import(self):
...
-        sync = self.binding_record.sync_date
+        sync = self.binding.sync_date
         if not sync:
             return True
-        record = self.backend_adapter.read(self.magento_id,
+        record = self.backend_adapter.read(self.external_id,
                                            attributes=['updated_at'])
         if not record['updated_at']:
             # in rare case it can be empty, in doubt, import it
             return False
```
lasley: Shouldn't this be `True`?

guewen: Seems so... according to the comment I'd say it should be True.

guewen:
Thanks @jaredkipe, you're helping me a lot! It turns out this test has been added here https://github.com/guewen/connector-magento/pull/6 so @Cedric-Pigeon might have his say on your proposal for the location field. I'll start by fixing the test so we can move forward :)

guewen:
The last commit should fix the test. I took the approach to change the location in the context of the binding, because it seems the test tries to check that we are exporting the value for "this" location. @Cedric-Pigeon your approval would be great :)

Thanks again for the help @jaredkipe

jaredkipe:
> Thanks @jaredkipe, you're helping me a lot! It turns out this test has been added here guewen#6 so @Cedric-Pigeon might have his say on your proposal for the location field. I'll start by fixing the test so we can move forward :)

@guewen You are very welcome, thank you for your fantastic work on all of these projects!

In my 'customizations' I actually added a location field to use like this.  This allowed me to 'group' several warehouses by injecting an intermediate 'view' location. This isn't ideal, I'd prefer to specify multiple locations or warehouses and put those into the context.  This should be possible as `stock.models.product.Product._get_domain_locations()` will use a list of location ids or a list of warehouse ids.  I'll get there eventually, I just have larger things to do.

jaredkipe:
@guewen I'm still being plagued by the imports eventually freezing up.  When I go to import 3k products, eventually the process will stop without explicitly failing.

Sometimes they fail with a result 'Could not acquire advisory lock'.  If this happens I need to restart odoo.

Here is a gist of active Postgres activity when I come back after trying to import stuff overnight. https://gist.github.com/jaredkipe/e81599fe5610e5a7f0eb59334d6529ed

If I find the 'started' jobs, I can requeue them, sometimes that gets them going again.. but only for so long.  I'm testing right now with various timeouts for transactions and and idle sessions in Postgres (9.6).

The same problem exists in threaded and worker modes.

guewen:
> I'm still being plagued by the imports eventually freezing up. When I go to import 3k products, eventually the process will stop without explicitly failing.

Do you have a magento test database dump to provide or is a private one?

guewen:
Can we merge?

jaredkipe:
> Do you have a magento test database dump to provide or is a private one?
It is private.  I've found that if I can work the list down more confidently by setting `lock_timeout` and `idle_in_transaction_session_timeout` to a couple of minutes, the whole process is much more reliable.  

An improvement that could be made to the item import would be to batch the products right after returning from the initial search of product ids as you must wait quite a while between the return (with thousands + product ids) to jobs actually starting to run.  (presumably because until the main transaction commits, the queue can't see them).

Bear in mind that the contention I have been dealing with is as it is importing individual products.  I have never had a problem with the initial import other than thinking it wasn't working while waiting for it to finish.

> Can we merge?

I think so.  I think all of the remaining issues I have will have to wait until later as things like the bind by SKU are now integral to my current workflow and I cannot stop to work on them.

jaredkipe:
@guewen I forgot to mention, the dependencies could use updating.  Here is the list of modules I needed to install the magento connector and the packages they come from (symlinks)

```
account_payment_mode -> ../oca/bank-payment/account_payment_mode
account_payment_partner -> ../oca/bank-payment/account_payment_partner
account_payment_sale -> ../oca/bank-payment/account_payment_sale
base_exception -> ../oca-beta/server-tools/base_exception
base_technical_user -> ../oca-beta/server-tools/base_technical_user
component -> ../oca-beta/connector/component
component_event -> ../oca-beta/connector/component_event
connector -> ../oca-beta/connector/connector
connector_base_product -> ../oca-beta/connector/connector_base_product
connector_ecommerce -> ../oca-beta/connector-ecommerce/connector_ecommerce
connector_magento -> ../oca-beta/connector-magento/connector_magento
dbfilter_from_header -> ../oca-server-tools/dbfilter_from_header
partner_gravatar -> ../hibou-gravatar-partner
product_multi_category -> ../oca/product-attribute/product_multi_category
queue_job -> ../oca/queue/queue_job
sale_automatic_workflow -> ../oca/sale-workflow/sale_automatic_workflow
sale_automatic_workflow_payment_mode -> ../oca/sale-workflow/sale_automatic_workflow_payment_mode
sale_exception -> ../oca/sale-workflow/sale_exception
```

All of those are out of beta now, and if memory serves, the only one I couldn't find a use for was the `e-commerce` module itself (`connector_ecommerce` for sure though).

jaredkipe:
~~I've been looking more into the hangups.  Basically the only commonality I'm ever able to see is that the product import binding will be created (logged from _create) but that product is not actually in the database. And then in Postgres there is an 'idle in transaction' with a query of one of the two of these.~~ https://gist.github.com/jaredkipe/1ab32a983e02c9ada64c21e90572f0a6

Okay.  After spending more hours working on this. It 'looked' like it was somewhere having to do with downloading the image right after the product is created.  It appears that CloudFlare was doing something to make that hang (no timeout?) in `_get_binary_image`. After hard coding the IP in /etc/hosts in the container suddenly I'm able to do thousands and thousands in the time it would have hung up on a couple.

This also explains why it was so degenerative, it seemed to only get worse and worse.  Well I guess because CloudFlare was limiting/throttling it more and more.

guewen:
@jaredkipe thanks for the details, this is very reassuring :) 

Could you retarget your pull requests on 10.0 now that's merged? So it'll ease the review.

jaredkipe:
@guewen Will do!  (I asked about that between getting the notice that this was merged and getting this comment. :D )

PR #259, opened 2017-09-13
[WIP] [10.0] connector_magento_payment_ref: Migration to 10.0

lmignon:
- [x] depends on #255 
- [ ] depends on https://github.com/OCA/sale-workflow/pull/526/


rousseldenis:
@lmignon https://github.com/OCA/sale-workflow/pull/526 is being merged

PR #264, opened 2017-10-10
Add configurable support

PierrickBrun:
Following the example of the bundle product, empty methods are created in order to be inherited.
That way the "flat" default system stays the same and the additional module will allow to import the configurable products in the Odoo format

- [x] import product attribute

- [x] import product attributes values

- [x] import product attribute lines

- [x] change product template

- [x] remove useless product templates
- [x] import product attribute prices
- [x] find a way to hide the principal product.product (linked to the magento configurable product) without removing the magento.product.product associated
example on the first line : configurable product.product
- [x] unit tests


bealdav:
Hi @gurneyalex 
Pierrick is a new talentuous kid on the akretion block (but completely new in odoo) . Could you add it in the akretion team/cla ? Thanks.

bealdav:
@guewen 
A said Pierrick'll contribute on Magento for configurable products.
He have to learn a lot here but he dive in your code for its first job in Odoo. It seems he like it
We now have needs to decide what is the way to deal with configurable products.

Our plan is (please fix us):
- import configurable as simple product : almost DONE
- convert this simple product to a full set of product variants (with attributes)
- inactive useless variant

The main question here is: step 2 and 3 should be synchro or deferred with a new job ?

cc @sebastienbeau

guewen:
> The main question here is: step 2 and 3 should be synchro or deferred with a new job ?

As step 1 alone doesn't make sense without 2 and 3, it seems to me that it can be synchronous.

connector_magento_configurable/models/inherit.py
```diff
@@ -0,0 +1,70 @@
...
+    _inherit = 'product.template'
+
+    @api.multi
+    @job
+    def delayable_unlink(self):
+        """
+            Allows to delay an unlink job.
+            This is used to delete the orphaned product templates
+        """
+        self.unlink()
```
guewen: A job should be idempotent: if run twice, should not fail. Here, you should just check if the product doesn't exist, ignore the unlink.

connector_magento_configurable/models/product_attribute_line/importer.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def run(self, magento_record, force=False):
+        self.magento_record = magento_record
+        super(ProductAttributeLineImporter, self).run(
+            magento_record['external_id'],
+            force,
+            )
+        self.env['product.template'].search([
+            ('product_variant_ids', '=', False)
+        ]).with_delay(priority=101, eta=60).delayable_unlink()
```
guewen: Is there a reason for giving an eta in 60 seconds?

guewen (commented):
A few details spotted on quick eyeballing

connector_magento_configurable/models/product_attribute_line/importer.py
PierrickBrun: The goal is to avoid trying to unlink a template still having product.product linked to it. Second thought, not sure if this is necessary but didn't think of another way to avoid the problem.

connector_magento_configurable/models/inherit.py
PierrickBrun: Obviously, yes

PierrickBrun:
Will start testing tomorrow, I think the functional perimeter is OK.

@guewen Do you have an idea how to hide the product.product without attributes ? (screenshot in the first post). If i disable it, the connector will stop updating the template.

Also, what is your opinion on the problem in product attribute prices ?
In magento, products created before a change in attribute prices don't change their price. In odoo they do -> can result in prices mismatch
Is an option to (de)activate it necessary or do we rely on sale order to handle this possible mismatch ?

guewen:
> @guewen Do you have an idea how to hide the product.product without attributes ? (screenshot in the first post). If i disable it, the connector will stop updating the template.

What is needed to update on the template? I'm really far to have the whole context, so I can be totally wrong, but I see these options:
* use another 'active' field on the binding which overrides the template's one
* consider even the inactive templates on the code that "update the template" (but I don't what you speak about)



> Also, what is your opinion on the problem in product attribute prices ?
> In magento, products created before a change in attribute prices don't change their price. In odoo they do -> can result in prices mismatch
> Is an option to (de)activate it necessary or do we rely on sale order to handle this possible mismatch ?

I don't know what this problem is... In any way there must be a check in the sale order (it exists I think).

guewen (approved):
Thanks!

guewen (changes requested):
Oops, I reviewed only the last commit. What's the status, still WIP?

connector_magento_configurable/models/configurable.py
```diff
@@ -0,0 +1,66 @@
...
+    to templated product, creates a job
+    """
+    _name = 'magento.product.configurable.batch.importer'
+    _inherit = 'magento.delayed.batch.importer'
+    _apply_on = ['magento.product.configurable']
+
+    def run(self, filters=None):
+        """ Run the synchronization """
+        from_date = filters.pop('from_date', None)
+        # to_date = filters.pop('to_date', None)
```
guewen: is it something which need to be done?

connector_magento_configurable/models/configurable.py
```diff
@@ -0,0 +1,66 @@
...
+        criterias = [('product_type', '=', 'configurable')]
+        if from_date:
+            criterias.append(('write_date', '>', from_date.strftime(
+                tools.misc.DEFAULT_SERVER_DATETIME_FORMAT)))
+        internal_ids = self.env['magento.product.product'].search(criterias)
```
guewen: preferable to give another name than something_ids when it holds records and not ids :)

connector_magento_configurable/models/configurable.py
```diff
@@ -0,0 +1,66 @@
...
+    _apply_on = ['magento.product.configurable']
+
+    def run(self, filters=None):
+        """ Run the synchronization """
+        from_date = filters.pop('from_date', None)
+        # to_date = filters.pop('to_date', None)
+        criterias = [('product_type', '=', 'configurable')]
+        if from_date:
+            criterias.append(('write_date', '>', from_date.strftime(
+                tools.misc.DEFAULT_SERVER_DATETIME_FORMAT)))
```
guewen: Instead of `strftime(tools.misc.DEFAULT_SERVER_DATETIME_FORMAT)` you should use `fields.DateTime.to_string()`

connector_magento_configurable/models/product_attribute_line/importer.py
```diff
@@ -0,0 +1,139 @@
...
+                                               attribute, value):
+        line = self._get_magento_product_attribute_line(
+            attribute,
+            value,
+            record
+        )
+        self._import_record(line)
+        self.env['product.template'].search([
+            ('product_variant_ids', '=', False)
+        ]).unlink()
```
guewen: It seems a bit risky to do that. How do you know you are deleting only what you should? There is for instances addons which changes the default behavior, variants are not created automatically and you have to create them with a button. Would be catastrophic here.

guewen (commented):
Looks good overall. Do you plan to add tests?

bealdav:
Yes @PierrickBrun planned to do some tests

PierrickBrun:
Thanks for the feedback, yes there will be tests

connector_magento_configurable/models/product_attribute_line/importer.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def attribute_code_field(self):
+        """ allows to override the field where the attribute_code is stored"""
+        return 'name'
+
+    def _write_product(self, magento_product, tmpl_id, value_ids):
+        magento_product.write(
+            {'product_tmpl_id': tmpl_id,
+             'attribute_value_ids': value_ids})
+        magento_product.odoo_id.write(
```
sebastienbeau: Not needed writing on magento_product will write on magento_product.odoo_id

PierrickBrun:
Hi @guewen, I think everything is OK now, can you confirm ?

PierrickBrun:
relaunching travis

connector_magento_configurable/models/product_product/importer.py
```diff
@@ -0,0 +1,24 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class ProductImporter(Component):
+    _inherit = 'magento.product.product.importer'
+
+    def _update(self, binding, data):
+        """ Update an OpenERP record """
+        if binding._name == 'magento.product.product':
+            if binding.product_tmpl_id.magento_bind_ids:
```
sebastienbeau: if binding.product_variant_count > 1: is better

PR #269, opened 2017-11-23
[10.0]Add new wizard to read data from magento

lmignon:
This wizard is useful when diagnosing a problem into the backend to get the json  returned by magento when you call the read method on a magento model.
The wizard is available on the backend.

### DIFF
--- a/connector_magento/__init__.py
+++ b/connector_magento/__init__.py
@@ -5,3 +5,4 @@
 
 from . import components
 from . import models
+from . import wizards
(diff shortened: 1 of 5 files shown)


guewen (approved):
Excellent! And really nice to see such a nice usage of the new component API :blush:

lmignon:
@guewen This component API is so powerful:smirk:

Cedric-Pigeon (approved):

PR #279, opened 2018-06-21
8.0 magento2

vimalvkg:


StefanRijnhart:
I think this branch needs some more work, and we're still waiting for the merge on the fulfilio lib. If you want to help, please consider fixing and adding tests.

IJOL:
JUst found that and badly needing it, so i will help, but tests and that s*** are not for me, anyway in your ( i assume )  expert position, where is that help needed? and sorry for be so late, just 9 days after you give up, is this heaven sign? ;)

thomaspaulb:
@StefanRijnhart When something is not finished, it at least needs some place where the code can be discussed, that is why normally a [WIP] pull request is opened. Is there any reason why you don't do this and others do?

For example now, there is a merge conflict.

StefanRijnhart:
@thomaspaulb I guess that's another way of looking at it. I'll open my own pull request if you think that will help.

StefanRijnhart:
Here you go: https://github.com/OCA/connector-magento/pull/289. Will have a look at the conflict later.

PR #296, opened 2019-08-23
[FIX] Negative paid amounts

StefanRijnhart:


StefanRijnhart:
Possibly due to customizations on the Magento side in our project, but we have encountered negative unpaid amounts in our order exports.

PR #299, opened 2020-02-17
[FIX] magento.website: map backend_id only on create

sebalix:
Without this fix when we synchronize the metadata of a backend it
triggers the recompute of 'magento.res.partner'.backend_id field
(related stored) because the 'backend_id' is updated on
'magento.website'.
So with a high number of bindings of partners in the database, this
recompute consume a huge amount of memory.

Also, a 'magento.website' should not change its backend with time, so
better to enforce this rule.

### DIFF
--- a/magentoerpconnect/magento_model.py
+++ b/magentoerpconnect/magento_model.py
@@ -635,6 +635,7 @@ class WebsiteImportMapper(ImportMapper):
         return {'name': name}
 
     @mapping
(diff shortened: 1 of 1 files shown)


guewen (approved):

guewen:
Travis fails on 7.0, fast-tracking this small fix.

StefanRijnhart (approved):

PR #292, opened 2019-04-19
[12.0][MIG] connector_magento

sebalix:
PR itself ready to be reviewed, but dependencies are not ready.

Depends on:

- <s>base_technical_user: https://github.com/OCA/server-tools/pull/1560</s>
- <s>connector_ecommerce: https://github.com/OCA/connector-ecommerce/pull/49</s>
    - <s>sale_exception: https://github.com/OCA/sale-workflow/pull/831</s>
        - <s>base_exception: https://github.com/OCA/server-tools/pull/1557</s>
    - <s>sale_automatic_workflow_payment_mode: https://github.com/OCA/sale-workflow/pull/786</s>
        - <s>sale_automatic_workflow: https://github.com/OCA/sale-workflow/pull/830</s>
        - <s>[BUG] account_payment_sale: https://github.com/OCA/bank-payment/issues/536</s>
          fixed by https://github.com/OCA/bank-payment/pull/630/files

`oca_dependencies.txt` has been updated to include all the dependencies and install the module on Travis.

<s>Documentation generation is stuck because of this bug: https://github.com/odoo/odoo/issues/28125
Don't know what to do about this, excepting disable the `api/api_models` part of the documentation (need to check if some other parts depend on it).</s>
EDIT: I disabled the documentation for class members using `fields` as a workaround.

PR #304, opened 2020-05-08
Add/12.0/magento2

StefanRijnhart:


StefanRijnhart:
Contains all of https://github.com/OCA/connector-magento/pull/292 until that one is merged

MPishpecki (approved):
Tested the configuration sync and order import. Magento 2.3.5 specifically.

PhilippeBach (approved):
Using this version in production since one month with odoo 12 and magento 2.3.4 
very good job

StefanRijnhart:
@PhilippeBach thanks for your feedback here but especially for reporting the bugs when I was still working on this!

pedrobaeza:
Should we merge this?

pedrobaeza:
(some commit squashing seems to be needed before)

StefanRijnhart:
If we accept this as superseding https://github.com/OCA/connector-magento/pull/292 (which it includes), then I can do some squashing.

StefanRijnhart:
@pedrobaeza I rebased but I guess it's a bit subjective which commits from the same authors to keep and which to squash so I hope it's ok.

pedrobaeza:
Yes, just a bit as you have done is enough, thanks

/ocabot merge nobump

PR #308, opened 2020-10-22
[12.0][ADD] Allow to import orders with a certain delay

StefanRijnhart:
This is useful when orders are imported frequently as it prevents orders
from being found unpaid if the transfer to status 'paid' is delayed at the
payment provider (set this value to the duration of that delay).

### DIFF
--- a/connector_magento/models/magento_storeview/common.py
+++ b/connector_magento/models/magento_storeview/common.py
@@ -85,9 +85,8 @@ class MagentoStoreview(models.Model):
             delayable = sale_binding_model.with_delay(priority=1)
             filters = {
                 'magento_storeview_id': storeview.external_id,
-                'from_date': (fields.Datetime.to_string(from_date)
-                              if from_date else None),
-                'to_date': fields.Datetime.to_string(to_date),
+                'from_date': from_date or None,
+                'to_date': to_date,
             }
             delayable.import_batch(backend, filters=filters)
             # Records from Magento are imported based on their `created_at`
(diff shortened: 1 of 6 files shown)


connector_magento/models/magento_storeview/common.py
```diff
@@ -74,32 +74,35 @@ def import_sale_orders(self):
...
+            if from_date and to_date < from_date:
+                continue
 
             delayable = sale_binding_model.with_delay(priority=1)
             filters = {
                 'magento_storeview_id': storeview.external_id,
-                'from_date': from_date,
-                'to_date': import_start_time,
+                'from_date': (fields.Datetime.to_string(from_date)
+                              if from_date else None),
```
hbrunn: if `False` is also fine, you don't need this check: https://github.com/OCA/OCB/blob/12.0/odoo/fields.py#L1790

hbrunn (changes requested):
I think it would be good to explain in the readme when to use the delay, and when to rely on `IMPORT_DELTA_BUFFER` - my assumption is you'll want the delay on a busy site?

connector_magento/models/magento_storeview/common.py
StefanRijnhart: Nooooo. I had troubles reproducing this, but I'll try again now.

StefanRijnhart: I don't understand my initial answer anymore, but the answer is that False is not a valid value for this key.

StefanRijnhart:
@hbrunn sorry for losing sight of this PR! I've documented this parameter now in the RST docs where I'm not mentioning `IMPORT_DELTA_BUFFER` as this is a technical and somewhat trivial qualification of the concept of 'now'.

hbrunn (approved):

StefanRijnhart:
@hbrunn thanks for the fix!

MPishpecki (approved):

PR #309, opened 2020-10-22
[12.0][IMP] Set a pricelist based on the currency of the order

StefanRijnhart:
The diff of some of the tests files shows that orders have been created in the wrong currency all along....


hbrunn (approved):
it put me off a bit that this is some arbitrary price list, but then when we import orders the price list isn't relevant for the prices anyways

guewen (approved):

PR #318, opened 2021-04-08
[10.0][MIG] connector_magento: rename columns

StefanRijnhart:


hbrunn (approved):
I think what you do in the body of the loop would be a nice addition to `rename_colums` with some parameter `ignore_missing` or similar, but that's out of scope here

Issue #319, opened 2021-04-08
Why commit "[IMP] Set a pricelist based on the currency of the order" fa27e776b093d3bc3a4998f2ee5a16084e964ef0 


IJOL:
This is a not starter for us, in our use case and AFAIK in the rest of the code is assumed that the currency of the order in magentio is ignored, and the import itself ( see lines 799-812 line price mapping ) assume the imported prices are in base currency, so why bother with pricelist currency at all, at least make this "Feature" ( for me is clearly a bug but who knows) optional, it's totally against our use case

PR #322, opened 2021-05-05
[12] Solved all showstopper bugs for importing from magento 1.X #319

IJOL:


StefanRijnhart:
Some good fixes here, and some we should discuss (as you can tell from the failing tests, it looks like some of the bugfixes should not apply to Magento2). Can you propose them one fix per PR? That would make it more manageable.

PR #326, opened 2021-08-22
[12.0][FIX] connector_magento: assign company from warehouse

StefanRijnhart:
The fiscal position is fetched in the sale order's onchange methods from the
partner properties based on the company of the sale order. But the onchanges
are called on a newId that does not have a company_id assigned yet.

Assign the company from the warehouse to allow for multicompany setups.


astirpe (approved):
Of course!

gurneyalex (approved):

PR #330, opened 2021-11-17
[FIX] connector_magento: check right key depending on Magento version

MPishpecki:


astirpe (approved):
:+1:  (assuming that Travis is green)

StefanRijnhart (approved):

